PR #16, opened 2020-06-15
Add wappalyzer attack module

bretfourbe:
Add wappalyzer attack module :  web technology detection (with versions and categories).
Fix setup.py, Pipefile, MANIFEST.in and attack.py to support the new mod_wapp module.
Fix undefined variable in crawler.py.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -5,6 +5,7 @@ include VERSION
 recursive-include wapitiCore/config *
 recursive-include wapitiCore/language_sources *
 recursive-include wapitiCore/report_template *
+recursive-include wapitiCore/wappalyzer/data *
 recursive-include doc *
 exclude wapitiCore/language_sources/*.sh
 exclude wapitiCore/language_sources/file_list.txt
--- a/Pipfile
+++ b/Pipfile
@@ -15,6 +15,8 @@ requests = ">=2.23.0"
 responses = "==0.10.6"
 tld = ">=0.11.11"
 yaswfp = ">=0.9.3"
+lxml = ">=4.5.0"
+six = ">=1.15.0"
 
 [requires]
 python_version = "3.8"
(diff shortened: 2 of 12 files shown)


fwininger:
You can not add code in GPLv3 because it's incompatible with the wapiti licence (GPLv2)

devl00p:
> You can not add code in GPLv3 because it's incompatible with the wapiti licence (GPLv2)

I'm not sure they are using GPLv3.

I saw a wappalyzer repo on github using GPLv3 but it seems to be a fork.

From the repository ( https://github.com/aliasio/wappalyzer ) I found from wappalyzer.com it seems that the official one is under MIT so it should be ok to use the json data (I saw no mentions about the json files)

devl00p:
Just make sure the apps.json comes from the repository under MIT

fwininger:
> 
> 
> Just make sure the apps.json comes from the repository under MIT

apps.json is under MIT.

The issue came from the .py file.

fwininger:
and actually we are sure if we want apps.json with 15 000 lines in the repos, because this file change often.

devl00p:
The json file may be dowloaded when the module is first launched, the same way the nikto module works.

Then we may consider adding a --update option to wapiti to sync those files.

Still the python code must comply with GPL v2

fwininger:
> The json file may be dowloaded when the module is first launched, the same way the nikto module works.

We don't want something like this. In our use case we a not able to have internet access at the first launch. We need to include everything in the installation process.

> Still the python code must comply with GPL v2

In progress

bretfourbe:
Hi ! Could you check this one is now ok ?

devl00p:
I'm merging this and I will push some changes on master after ad I'm not yet familiar with pull requests. I made some changes to use Wapiti Page object

PR #17, opened 2020-06-16
Add a new report section "Additionals"

bretfourbe:
Add a new Additional class in language/vulnerability.py, a new xml parser in file/additionalxmlparser.py and a new xml file in config/vulnerabilities/additionals.xml.
Add new methods add_additional and add_additional_type in attack.py, reportgenerator.py and sqlite_persister.py and call them in wapiti.py.
Fix report generator files (execpt vulneranetxml) and report.html template.
Add the necessary translations in en.po and fr.po and generate again the *.mo files.
Fix attack tests (FakePersister) to support these changes.


devl00p:
Nice and useful :) Than you

PR #22, opened 2020-07-22
Add check cookie module

OussamaBeng:
Adding new module : Check if the "Secure" and "HttpOnly" flags are set in cookies


fwininger:
@OussamaBeng why you have the commit "Add Brute Force Basic HTTP Authentication module" in our PR ? Can you clean this, please

PR #27, opened 2020-08-04
Add module csrf

bretfourbe:
Add module csrf, update attack.py, vulnerabilities.xml and vulnerability.py consequently, add translations.


wapitiCore/attack/mod_csrf.py
```diff
@@ -0,0 +1,203 @@
...
+from wapitiCore.net.crawler import Page
+
+class mod_csrf(Attack):
+    """
+    This class implements a CSRF/XSRF attack
+    """
+
+    name = "csrf"
+
+    do_get = True
```
devl00p, line 18: I guess we can set `do_get` to False

wapitiCore/attack/mod_csrf.py
```diff
@@ -0,0 +1,203 @@
...
+    do_post = True
+
+    csrf_string = None
+
+    MIN_ENTROPY = 3
+    MIN_LENGTH = 8
+
+    TOKEN_FORM_STRINGS = [
```
devl00p, line 26: I don't know what is the "academic" way to check for CSRF tokens but known names seems good.

Using entropy() on form fields (names) may be a way to discover dynamic CSRF tokens but there will be risk of more false positives of course.

wapitiCore/attack/mod_csrf.py
```diff
@@ -0,0 +1,203 @@
...
+
+        # csrf value entropy is too low, i.e. csrf value is too predictive
+        if self.entropy(csrf_value) < self.MIN_ENTROPY:
+            return False
+
+        return True
+
+    @staticmethod
+    def is_same_response(original_response: Page, mutated_response: Page):
+        """Check whether mutated_response status and content are identical to those of original_response"""
```
devl00p, line 77: I'm a bit concerned with that method but I think it is impossible to have perfect results for this module.

The way Wapiti crawls a website may result in a token being changed before the form is submitted if the website change the token after each request so there may be cases where the `original_response` informations aren't useful.

wapitiCore/attack/mod_csrf.py
bretfourbe: Known names seems ok on websites developed with known frameworks but it is limited since it is recommended to use custom names for csrf token fields. 

Several tokens won't be detected as it is, so many false positives will be raised. I don't know how to ensure that a field is a csrf token when it is a custom name. I don't understand how entropy() could help on that.

wapitiCore/attack/mod_csrf.py
bretfourbe: > I'm a bit concerned with that method but I think it is impossible to have perfect results for this module.

This is a first step to compare the website behavior when sending valid and invalid csrf tokens, but i'm aware that comparing response contents might require to use something like a distance calculation, in order to take in account little variations (new csrf token value for instance). However it won't raise false positives as it is.

> The way Wapiti crawls a website may result in a token being changed before the form is submitted if the website change the token after each request so there may be cases where the original_response informations aren't useful.

Is this linked to #31 ?

devl00p: For the time being comparing the response should be enough.

Wapiti use a FIFO queue to crawl URLs discovered so if it crawls a web page with a form this one will be at the end of the queue and other webpages (those at the same "depth" that the current one) will be processed before the form.

It may have an impact on the token if the website is changing the token at every webpage fetched by the visitor... it depends on the way the website implement this

wapitiCore/attack/mod_csrf.py
```diff
@@ -0,0 +1,203 @@
...
+            if self.verbose >= 1:
+                print("[+] {}".format(original_request))
+
+            csrf_value = self.is_csrf_present(original_request)
+            # check if token is present
+            if not csrf_value:
+                self.log_red("---")
+                self.log_red(
+                    Vulnerability.CSRF + Vulnerability.MSG_FROM,
+                    original_request
```
devl00p, line 169: You should use the `.http_repr()` attribute so it prints nicely

wapitiCore/attack/mod_csrf.py
```diff
@@ -0,0 +1,203 @@
...
+        forms = self.persister.get_forms(attack_module=self.name) if self.do_post else []
+
+        for original_request in forms:
+
+            if self.verbose >= 1:
+                print("[+] {}".format(original_request))
+
+            csrf_value = self.is_csrf_present(original_request)
+            # check if token is present
+            if not csrf_value:
```
devl00p, line 165: This is very noisy as it will print something for every possible value of the form (see example with Acunetix website below)

We should reduce to one message `per original_request.url` / `original_request.post_keys` tuple

```
---
Falsification de requête inter-site (CSRF)  en provanance de POST http://testphp.vulnweb.com/cart.php (4)
        data: price=986&addcart=3
[+] POST http://testphp.vulnweb.com/cart.php (4)
        data: price=500&addcart=1
---
Falsification de requête inter-site (CSRF)  en provanance de POST http://testphp.vulnweb.com/cart.php (4)
        data: price=500&addcart=1
[+] POST http://testphp.vulnweb.com/cart.php (4)
        data: price=460&addcart=5
---
Falsification de requête inter-site (CSRF)  en provanance de POST http://testphp.vulnweb.com/cart.php (4)
        data: price=460&addcart=5
[+] POST http://testphp.vulnweb.com/cart.php (4)
        data: price=15000&addcart=7
---
Falsification de requête inter-site (CSRF)  en provanance de POST http://testphp.vulnweb.com/cart.php (4)
        data: price=15000&addcart=7
[+] POST http://testphp.vulnweb.com/cart.php (4)
        data: price=1000&addcart=4
---
Falsification de requête inter-site (CSRF)  en provanance de POST http://testphp.vulnweb.com/cart.php (4)
        data: price=1000&addcart=4
[+] POST http://testphp.vulnweb.com/cart.php (4)
        data: price=800&addcart=2
---
Falsification de requête inter-site (CSRF)  en provanance de POST http://testphp.vulnweb.com/cart.php (4)
        data: price=800&addcart=2
[+] POST http://testphp.vulnweb.com/cart.php (4)
        data: price=10000&addcart=6
---
```

wapitiCore/attack/mod_csrf.py
```diff
@@ -0,0 +1,203 @@
...
+                level=Anomaly.MEDIUM_LEVEL,
+                request=mutated_request,
+                info=anom_msg,
+            )
+
+        else:
+            if mutated_response.status == 500:
+                anom_msg = Anomaly.MSG_PARAM_500.format(self.csrf_string)
+
+                self.add_anom(
```
devl00p, line 136: We don't necessary need to check for 500 error here. It is more useful when we are using an aggresive payload (like those in sql / exec / file modules)

devl00p (changes requested):

PR #28, opened 2020-08-05
Add module : Checking CSP

OussamaBeng:
Add module to check basic CSP configuration


wapitiCore/language_sources/fr.po
```diff
@@ -732,6 +736,18 @@ msgstr "L'administrateur web doit supprimer manuellement les sauvegardes présen
...
+msgstr "Configuration du CSP"
+
+#: ../language/vulnerability.py:115
+msgid "CSP \"{0}\" is missing"
+msgstr "CSP \"{0}\" est manquant"
+
+#: ../language/vulnerability.py:116
+msgid "CSP \"{0}\" value is not safe"
+msgstr "La valeur du CSP \"{0}\" n'est pas sécurisée"
+
```
bretfourbe, line 750: Translations for Content Security Policy description and solution are missing in fr.po and en.po

wapitiCore/attack/mod_csp.py
```diff
@@ -0,0 +1,77 @@
...
+            if any(el in request.headers['Content-Security-Policy'] for el in check_list):
+                return 1
+        # If the tested element is none of the previous list, we must ensure that one of this safe values is present in the CSP
+        else:
+            if any(elm in request.headers['Content-Security-Policy'] for elm in check_list):
+                return 2
+            else:
+                return 1
+
+        return 2
```
bretfourbe, line 47: Nice refacto :)

wapitiCore/attack/mod_csp.py
```diff
@@ -0,0 +1,77 @@
...
+            return False
+        else:
+            return re.search(csp_attribute, request.headers['Content-Security-Policy'])
+
+    def check_default_src(self, request, element, check_list):
+        '''
+        This function return a number who tell us the status of the tested element in the CSP
+        if the function return -1 : the element is missing in the CSP
+        if the function return 1  : the element is set, but his value is not secure
+        if the function return 2  : the element is set and his value is secure
```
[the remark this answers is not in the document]
OussamaBeng: Fixed

wapitiCore/attack/mod_csp.py
```diff
@@ -0,0 +1,77 @@
...
+            result = self.check_default_src(response, def_src, self.all_check[i])
+            if result == -1:
+                self.log_red(Additional.MSG_CSP_MISSIING.format(def_src))
+                self.add_addition(
+                    category=Additional.INFO_CSP,
+                    level=Additional.LOW_LEVEL,
+                    request=request,
+                    info=Additional.MSG_CSP_MISSIING.format(def_src)
+                )
+            elif result == 1:
```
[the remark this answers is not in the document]
OussamaBeng: Fixed

wapitiCore/language_sources/fr.po
```diff
@@ -294,6 +294,10 @@ msgstr "(QUERY_STRING)"
 msgid "Error: The server did not understand this request"
 msgstr "Erreur : Le serveur n'a pas compris la requête"
 
+#: ../attack/mod_csp.py:122
+msgid "Checking CSP :"
+msgstr "Vérification du CSP :"
+
```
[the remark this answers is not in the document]
OussamaBeng: Fixed

wapitiCore/language_sources/fr.po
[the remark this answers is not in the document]
OussamaBeng: Fixed

OussamaBeng: Fixed

wapitiCore/attack/mod_csp.py
OussamaBeng: Thanks :)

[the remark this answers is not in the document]
OussamaBeng: Fixed

bretfourbe:
Seems good to me :) @devl00p ?

devl00p:
> Seems good to me :) @devl00p ?

I need to freshen my knowledge on the subject to be sure I'm not missing something but it is definitively on my list :)

wapitiCore/attack/mod_csp.py
```diff
@@ -0,0 +1,77 @@
...
+    check_object_src = ["object-src 'none'"]
+    check_base_uri = ["base-uri 'none'", "base-uri 'self'"]
+
+    all_check = [check_default, check_script_src, check_object_src, check_base_uri]
+
+    def is_set(self, request: object, csp_attribute):
+        if 'Content-Security-Policy' not in request.headers:
+            return False
+        else:
+            return re.search(csp_attribute, request.headers['Content-Security-Policy'])
```
devl00p, line 24: As you are not giving regex object but just string using `in` should be enough

wapitiCore/attack/mod_csp.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+# This module check the basics recommendations of CSP
+class mod_csp(Attack):
+    name = "csp"
+
+    checks = ["default-src", "script-src", "object-src", "base-uri"]
+
+    check_default = ["default-src *", "default-src 'unsafe-eval'"]
+    check_script_src = ["script-src 'unsafe-inline'", "script-src 'data:'", "script-src 'http:'", "script-src 'https:'"]
```
devl00p, line 14: Using the policy prefix  each time you won't match values when they aren't in the first position.

Such an example is shown here :
https://developers.google.com/web/fundamentals/security/csp/#use_case_3_ssl_only

`script-src https: 'unsafe-inline';`

Here `unsafe-inline` is in second position but it won't be detected.

Also it seems that `http:` or `https:` or `https://*.example.com` or `http://example.com` doesn't need to be put between quotes.

It should be possible to use a regex (or a very basic parser) to get the several values of the policy.

wapitiCore/attack/mod_csp.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def attack(self):
+        url = self.persister.get_root_url()
+        request = Request(url)
+        response = self.crawler.get(request, follow_redirects=True)
+
+        self.log_blue(_("Checking CSP :"))
+        print("CSP : ", response.headers['Content-Security-Policy'])
+        i = 0
+        for def_src in self.checks:
```
devl00p, line 57: As CSP policies are separated by semi-colon the first thing to do should be to split them on that character and then separate policy name from values with the colon.

devl00p (changes requested):

wapitiCore/attack/mod_csp.py
devl00p: Here is an example : 

```
import re

regex = re.compile(r"\s*((?:'[^']*')|(?:[^'\s]+))\s*")
csp = "default-src https: https://cdn.example.net; script-src https: 'unsafe-inline' 'nonce-EDNnf03nceIOfn39fn3e9h3sdfa' 'sha256-qznLcsROx4GACP2dm0UCKCzCG-HiZ1guq6ZZDob_Tng='; style-src https: 'unsafe-inline'"

for policy_string in csp.split(";"):
    policy_name, policy_values = policy_string.strip().split(" ", 1)
    print(policy_name)
    print("="*20)
    for value in regex.findall(policy_values):
        print(value.strip("'"))
    print('')
```

output : 

```
default-src
====================
https:
https://cdn.example.net

script-src
====================
https:
unsafe-inline
nonce-EDNnf03nceIOfn39fn3e9h3sdfa
sha256-qznLcsROx4GACP2dm0UCKCzCG-HiZ1guq6ZZDob_Tng=

style-src
====================
https:
unsafe-inline
```

wapitiCore/attack/mod_csp.py
```diff
@@ -0,0 +1,97 @@
+import requests, re
```
devl00p, line 1: You should respect import order convention https://stackoverflow.com/questions/22722976/import-order-coding-standard/22789161#22789161

wapitiCore/attack/mod_csp.py
```diff
@@ -0,0 +1,102 @@
...
+    check_list = ["default-src", "script-src", "object-src", "base-uri"]
+
+    check_default = ["*", "unsafe-eval"]
+    check_script_src = ["unsafe-inline", "data:", "http", "https"]
+    check_object_src = ["none"]
+    check_base_uri = ["none", "self"]
+
+    all_check = [check_default, check_script_src, check_object_src, check_base_uri]
+
+    def check_policy_values(self, policy_name, check_list, csp_dict):
```
devl00p, line 21: Prototype of this function is name because it will be easy to write tests (and we need them)

wapitiCore/attack/mod_csp.py
```diff
@@ -0,0 +1,102 @@
...
+            regex = re.compile(r"\s*((?:'[^']*')|(?:[^'\s]+))\s*")
+            for policy_string in csp.split(";"):
+                try:
+                    policy_name, policy_values = policy_string.strip().split(" ", 1)
+                except ValueError:
+                    # Either it is malformed or we reach the end
+                    continue
+                csp_dict[policy_name] = [value.strip("'") for value in regex.findall(policy_values)]
+
+            self.log_blue(_("Checking CSP :"))
```
devl00p, line 78: In order to comply with other modules behavior I guess we will remove this log and only print useful information. Most modules only print log when something interesting is found.

wapitiCore/attack/mod_csp.py
```diff
@@ -0,0 +1,102 @@
...
+        0  : the element is set, but his value is not secure
+        1  : the element is set and his value is secure
+        """
+
+        if policy_name not in csp_dict and "default-src" not in csp_dict:
+            return -1
+
+        # The HTTP CSP "default-src" directive serves as a fallback for the other CSP fetch directives.
+        # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/default-src
+        policy_values = csp_dict.get(policy_name, csp_dict["default-src"])
```
devl00p, line 34: Would not have guessed Python could raise a KeyError at this point ^_^

PR #29, opened 2020-08-07
Add post authentication using existing options in command line

bretfourbe:
Using --auth-type=post -a=<username>%<password> -s <login_url> the crawler will start trying to login with the creds provided after having retrieved the form <login_url>.
Add new method in Page class which look for a login form.
Add new method in Crawler class that try to login.
Add verify parameter to get and request in Crawler to allow secure (or not) connections and so avoid SSL errors.


wapitiCore/main/wapiti.py
```diff
@@ -1150,8 +1153,13 @@ def wapiti_main():
                     pass
                 else:
                     print(_("[*] Resuming scan from previous session, please wait"))
+
+                    if "auth_type" in args and args.auth_type == "post":
+                        wap.crawler.try_login(wap.crawler.auth_url)
                     wap.browse()
             else:
+                if "auth_type" in args and args.auth_type == "post":
```
jinxka, line 1161: there should be a way to avoid this repetition

wapitiCore/main/wapiti.py
```diff
@@ -1057,7 +1057,10 @@ def wapiti_main():
                 raise InvalidOptionValue("-a", args.credentials)
 
         if "auth_type" in args:
-            wap.set_auth_type(args.auth_type)
+            if args.auth_type == "post" and args.starting_urls != []:
+                wap.crawler.auth_url = args.starting_urls[0]
```
devl00p, line 1061: Nice idea to use `-s` :) We will have to document it though

wapitiCore/net/crawler.py
```diff
@@ -1154,6 +1202,58 @@ def auth_method(self, value):
             from requests_kerberos import HTTPKerberosAuth
             self._session.auth = HTTPKerberosAuth()
 
+    def try_login(self, auth_url: str):
+        """Try to authenticate with the provided url and credentials."""
+        if len(self._auth_credentials) != 2:
+            print(_("Login failed"))
```
devl00p, line 1208: We should be a little more verbose (`Invalid credentials format` maybe ?)

wapitiCore/net/crawler.py
```diff
@@ -1154,6 +1202,58 @@ def auth_method(self, value):
...
+            return
+
+        username, password = self._auth_credentials
+        headers = {
+            'User-Agent': self.user_agent,
+            'Accept': 'text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8',
```
devl00p, line 1214: Those `Accept*` headers should be standard for the crawler session if not already present.

I saw some case where an XSS won't happen if html is not mentionned in `Accept`

wapitiCore/net/crawler.py
```diff
@@ -926,6 +926,52 @@ def iter_forms(self, autofill=True):
...
+                    password_field_idx.append(i)
+
+            # ensure login form
+            if len(username_field_idx) == 1 and len(password_field_idx) == 1:
+                inputs = form.find_all("input", attrs={"name": True})
+
+                url = self.make_absolute(form.attrs.get("action", "").strip() or self._url)
+                method = "POST"
+                enctype = form.attrs.get("enctype", "application/x-www-form-urlencoded").lower()
+                post_params = [[input["name"], input.get("value", "")] for input in inputs]
```
devl00p, line 959: You should use another name than `input` (standard python function name)

wapitiCore/net/crawler.py
```diff
@@ -926,6 +926,52 @@ def iter_forms(self, autofill=True):
...
+
+                elif input_type == "password":
+                    password_field_idx.append(i)
+
+            # ensure login form
+            if len(username_field_idx) == 1 and len(password_field_idx) == 1:
+                inputs = form.find_all("input", attrs={"name": True})
+
+                url = self.make_absolute(form.attrs.get("action", "").strip() or self._url)
+                method = "POST"
```
devl00p, line 957: A form may use GET. In fact GET is the default method if not specified in the form `method` attribute.

It may seems stupid to deal with those cases as `-s` can be leveraged for such cases (like `-s http://target/login.php?user=test&pass=test`) but it may be useful to deal with it it the form have some CSRF tokens

wapitiCore/net/crawler.py
```diff
@@ -926,6 +926,52 @@ def iter_forms(self, autofill=True):
...
                 yield new_form
 
+    def find_login_form(self):
+        """Returns the login Request extracted from the Page, the username and password fields."""
+
+        for form in self.soup.find_all("form"):
+            username_field_idx = []
+            password_field_idx = []
+
+            for i, input_field in enumerate(form.find_all("input")):
```
devl00p, line 936: But you should either make sure input type is text for username field or discard other common types.

For example I tried on a phpBB forum where two fields were incorrectly threated as login fields : 

```
input type=checkbox name=autologin
input type=submit name=login
```

wapitiCore/net/crawler.py
```diff
@@ -1172,7 +1272,8 @@ def get(self, resource: web.Request, follow_redirects: bool = False, headers: di
                 resource.url,
                 timeout=self._timeout,
                 allow_redirects=follow_redirects,
-                headers=headers
+                headers=headers,
+                verify=self.secure
```
devl00p, line 1276: Nice

wapitiCore/net/crawler.py
```diff
@@ -1154,6 +1202,58 @@ def auth_method(self, value):
...
+                    referer=login_form.referer,
+                    link_depth=login_form.link_depth
+                )
+
+                login_response = self.post(
+                    login_request,
+                    follow_redirects=True
+                )
+                # ensure logged in
+                if login_response.soup.findAll(text=re.compile(r'(?i)((log|sign)\s?out|disconnect|déconnexion)')) != []:
```
devl00p, line 1242: Tested on only two websites but it worked :)

devl00p (changes requested):
Good job :+1:

wapitiCore/net/crawler.py
bretfourbe: Added in Crawler __init__ method

wapitiCore/main/wapiti.py
bretfourbe: Thanks :) I will prepare a PR to update the documentation

PR #30, opened 2020-08-11
Add Authenticated Scan mode

OussamaBeng:
The user can add the credentials in the argument of wapiti for an authenticated scan
example : Wapiti -u http://example.com --auth-type=post -a login%password

### DIFF
--- a/wapitiCore/net/crawler.py
+++ b/wapitiCore/net/crawler.py
@@ -34,13 +34,13 @@ import pickle
 import math
 import functools
 from time import sleep
-
 # Third-parties
 import requests
 from requests.packages.urllib3 import disable_warnings
 from requests.packages.urllib3.exceptions import ReadTimeoutError
 from requests.exceptions import ConnectionError, RequestException, ReadTimeout, SSLError
 from requests.models import Response
+from requests import cookies
 from tld import get_fld
 from tld.exceptions import TldDomainNotFound, TldBadUrl
 from bs4 import BeautifulSoup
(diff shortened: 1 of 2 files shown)


wapitiCore/net/crawler.py
```diff
@@ -1016,6 +1017,93 @@ def __init__(
...
+
+        else:
+
+            page = self.get(web.Request(url), follow_redirects=True, headers=head, verify=False)
+            forms = []
+
+            for i, form in enumerate(page.iter_forms(autofill=False)):
+                if i == 0:
+                    print('')
+                    print(_("Choose the form you want to use or enter 'q' to leave :"))
```
devl00p, line 1063: This part works well as we are asking user input but see my comment on `set_session`

wapitiCore/net/crawler.py
```diff
@@ -1016,6 +1017,93 @@ def __init__(
...
+
+                    request = Request(page.url, post_params=post_params, link_depth=form.link_depth)
+                    response = self.post(request, follow_redirects=True)
+
+        return response
+
+    def set_session(self):
+
+        url = self._base.url
+        res = self.get_auth_form(url)
```
devl00p, line 1103: I wasn't able to make it work on `testphp.vulnweb.com` without modifying the code as

- the website use `uname` for username field and `pass` for password field (with -a `test%test` option)
- without specifying `-a` it find the correct form and fill it correctly but you are giving no choice for the base URL that is used to define the scope of the scanner so it won't do the scan if the user specify a full login url with the default scope (`folder`). This is the case with the acunetix website using `http://testphp.vulnweb.com/login.php` as login page.

wapitiCore/net/crawler.py
```diff
@@ -1016,6 +1017,93 @@ def __init__(
...
+        self.session_cookies =  requests.cookies.merge_cookies(res.cookies,self.session_cookies)
```
devl00p, line 1104: I didn't remember seing this `merge` function, interesting :)

OussamaBeng:
PR related to https://github.com/wapiti-scanner/wapiti/pull/29
Already merged

PR #36, opened 2020-09-14
Edit the payload of the  SQL Injection mod

OussamaBeng:
adding a "PAYLOADS_FILE" in the SQL Injection mod

### DIFF
--- a/wapitiCore/attack/mod_sql.py
+++ b/wapitiCore/attack/mod_sql.py
@@ -29,10 +29,15 @@ class mod_sql(Attack):
     """
     This class implements an error-based SQL Injection attack
     """
-
     time_to_sleep = 6
     name = "sql"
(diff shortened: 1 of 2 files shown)


wapitiCore/attack/mod_sql.py
```diff
@@ -32,7 +32,7 @@ class mod_sql(Attack):
 
     time_to_sleep = 6
     name = "sql"
-    payloads = ("\xBF'\"(", Flags())
+    PAYLOADS_FILE = "SQLPayloads.txt"
```
tarraschk, line 35: Ajouter la source du fichier et préciser la licence

devl00p:
I'm closing this pull request, the current version on the module relies on error messages only (heuristics) and those payloads are tied to different techniques.

I will work on boolean-based technique not for next release but the following one

PR #38, opened 2020-12-07
Fix pylint errors

fwininger:
This PR resolve the pylint errors

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,7 +20,7 @@ jobs:
           pip install pylint
       - name: Pylint
         run: |
-          pylint --rcfile=.pylintrc *.py wapitiCore tests
+          pylint --rcfile=.pylintrc *.py wapitiCore
 
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -20,7 +20,7 @@ jobs:
           pip install pylint
       - name: Pylint
         run: |
-          pylint --rcfile=.pylintrc *.py wapitiCore tests
+          pylint --rcfile=.pylintrc *.py wapitiCore
```
devl00p, line 23: Nice

.pylintrc
```diff
@@ -24,7 +24,9 @@ confidence=
 # --enable=similarities". If you want to run only the classes checker, but have
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
-disable=too-few-public-methods,too-many-instance-attributes,too-many-arguments,too-many-locals,logging-format-interpolation,not-an-iterable,too-many-public-methods,missing-function-docstring,missing-class-docstring,missing-module-docstring
+disable=too-few-public-methods,too-many-instance-attributes,too-many-arguments,
```
devl00p, line 27: That would be ok even if we should work on improving that :)

devl00p (approved):

PR #40, opened 2020-12-16
Edit Wappalyzer Output

OussamaBeng:
add the columns "app name", "app version", "app categories" to the "add_additional" output


wapitiCore/report/jsonreportgenerator.py
```diff
@@ -134,7 +134,10 @@ def add_additional(self, category=None, level=0, request=None, parameter="", inf
             "level": level,
             "parameter": parameter,
             "http_request": request.http_repr(left_margin=""),
-            "curl_command": request.curl_repr
+            "curl_command": request.curl_repr,
+            "app_name": "",
```
jinxka, line 138: Why do we send nothing here?

PR #45, opened 2021-01-18
Wapiti should accept urls without path

jinxka:
We should be able to use wapiti even when the URL has no given path.

I am not sure if this is the best way to set the URL path, if you know a better one, please let me know.


fwininger:
Thanks !

devl00p:
Seems ok to me, appending / when parts.path is empty should be enough, can't imagine breaking situations

PR #49, opened 2021-02-03
Add arg '--max-attack-time'

### LINKED ISSUE #48: Max attack time
Hello,

I am trying to automate the use of Wapiti in my project, but I have a problem with the duration of some modules.

I tried to use the '--max-scan-time' parameter but it only influences the browsing time.

There should be a way to limit the maximum duration of a module like '--max-scan-time' does for the browser

If you have an idea about this subject, please let know.

Thank you

jinxka:
Hello,

First, I changed the way '--max-scan-time' works as I don't understand why we apply a '*60' on the argument to force it to be in minutes.

'--max-scan-time' is now in seconds.

Then I added a new arg '--max-attack-time' which cause the module to stop if it's duration is superior.

The good thing is that the vulnerabilities found are not lost and the execution continue normally (the next module is executed with the same 'timeout' constraint)

wapiti-scanner/wapiti-scanner.github.io#1

Close #48

### DIFF
--- a/wapitiCore/main/wapiti.py
+++ b/wapitiCore/main/wapiti.py
@@ -581,8 +581,8 @@ class Wapiti:
     def set_scan_force(self, force: str):
         self._scan_force = force
 
-    def set_max_scan_time(self, minutes: float):
-        self._max_scan_time = minutes * 60
+    def set_max_scan_time(self, seconds: float):
+        self._max_scan_time = seconds
 
     def set_color(self):
         """Put colors in the console output (terminal must support colors)"""
@@ -881,8 +881,10 @@ def wapiti_main():
 
     parser.add_argument(
         "--max-scan-time",
-        metavar="MINUTES",
-        help=_("Set how many minutes you want the scan to last (floats accepted)"),
+        metavar="SECONDS",
+        help=_("Set how many seconds you want the scan to last (floats accepted)"),
+        type=float, default=0
+    )
         type=float, default=0
     )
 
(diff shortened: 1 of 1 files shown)


wapitiCore/main/wapiti.py
```diff
@@ -876,8 +885,15 @@ def wapiti_main():
...
-        help=_("Set how many minutes you want the scan to last (floats accepted)"),
+        metavar="SECONDS",
+        help=_("Set how many seconds you want the scan to last (floats accepted)"),
+        type=float, default=0
+    )
+
+    parser.add_argument(
+        "--max-attack-time",
+        metavar="SECONDS",
+        help=_("Set how many seconds you want the attack to last (floats accepted)"),
```
devl00p, line 896: You should mention that this is the time for each module and not for the whole attack process.

devl00p: Otherwise it seems good to me

devl00p (changes requested):

fwininger:
https://github.com/wapiti-scanner/wapiti/pull/50

wapitiCore/main/wapiti.py
jinxka: I changed it to "Set how many seconds you want each attack mdoule to last (floats accepted)"

jinxka: I also added a PR for the wapiti man https://github.com/wapiti-scanner/wapiti-scanner.github.io/pull/1

PR #53, opened 2021-02-04
Refactor url handling

jinxka:
I'm trying to normalize the URLs handling because I noticed that like 'URL' before the endpoints arguments does not accept an empty path.

I created a new method 'set_url_path' which add '/' to the url if no path has been set.

I also modified the is_valid_endpoint to is_valid_url which is a method that verify the format of an url.

### DIFF
--- a/wapitiCore/main/wapiti.py
+++ b/wapitiCore/main/wapiti.py
@@ -648,16 +648,25 @@ class Wapiti:
         return self.persister.have_attacks_started()
 
 
-def is_valid_endpoint(url):
+def fix_url_path(url):
+    """Fix the url path if its not defined"""
+    return url if urlparse(url).path else url + '/'
+
+
+def is_valid_endpoint(url_type, url):
+    """Verify if the url provided has the right format"""
     try:
         parts = urlparse(url)
     except ValueError:
+        print('ValueError')
         return False
     else:
         if parts.params or parts.query or parts.fragment:
+            print(_("Error: {} must not contain params, query or fragment!").format(url_type))
             return False
-        if parts.scheme in ("http", "https") and parts.netloc and parts.path:
+        if parts.scheme in ("http", "https") and parts.netloc:
             return True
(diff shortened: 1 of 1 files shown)


wapitiCore/main/wapiti.py
```diff
@@ -648,16 +648,25 @@ def have_attacks_started(self) -> bool:
...
+def is_valid_url(url_type, url):
+    """Verify if the url provided has the right format"""
     try:
         parts = urlparse(url)
     except ValueError:
+        print('ValueError')
         return False
     else:
         if parts.params or parts.query or parts.fragment:
+            print(_("Error: {} must not contain params, query or fragment!").format(url_type))
```
devl00p, line 665: This is not good for the URL given to wapiti using -u. You can provide an URL containing parameters if you are using the "url" scope

wapitiCore/main/wapiti.py
```diff
@@ -1034,15 +1043,9 @@ def wapiti_main():
...
+    url = set_url_path(args.base_url)
 
-    parts = urlparse(url)
-    # Set path to '/' if path is Null
-    if not parts.path:
-        url += '/'
-
-    if not parts.scheme or not parts.netloc:
-        print(_("Invalid base URL was specified, please give a complete URL with protocol scheme."))
+    if not is_valid_url('URL', url):
```
devl00p, line 1048: see first comment

wapitiCore/main/wapiti.py
```diff
@@ -1034,15 +1043,9 @@ def wapiti_main():
             print('')
         sys.exit()
 
-    url = args.base_url
+    url = set_url_path(args.base_url)
```
devl00p, line 1046: what about  `fix_url_path` ?

devl00p (changes requested):

wapitiCore/main/wapiti.py
jinxka: I reverted the changes for the main *url*

PR #54, opened 2021-02-05
Fix wappalyzer to support new format in apps.json

bretfourbe:
Change normalization method in wappalyzer to support generic format in apps.json.
For "cookies", "meta" and "headers", we now support the following list format :
`{"meta" : {"generator": ["test1", "test2"]}}` instead of only `{"meta" : {"generator": "test1"}}`


fwininger:
Support for this issue : https://github.com/AliasIO/wappalyzer/issues/3748

PR #58, opened 2021-02-06
Remove vulnerability descriptions from XML files + remove corresponding xml parsers

### LINKED ISSUE #57: Remove vulnerability descriptions from XML files
Vulnerability/Anomaly/Additional mechanism was added some times ago in a java-ish way, requiring a XML parser for each category to be loaded.

It has no real benefit as adding a a new catagory will require to add the code along with the parser and the xml file while it would be simpler with Python to just put information somewhere in the code and reference it later.

The other big issue is that the generateSources script can't see strings within those XML files so we had to put some kind of placeholders and put those placeholder back so they can be registered by the script.

This is bad because that placeholder appears in the translation file (.po) and people not knowing the project well may try to translate the message litteraly.

Let's refactor this and remove the XML files along with their parsers. Next step should be reports.

devl00p:
Fix #57


fwininger:
Looks good to me.

PR #64, opened 2021-02-10
Fix endpoint issue in ssrf module

### LINKED ISSUE #63: Bug: mod_ssrf doesn't use custom external_endpoint
Hello,

The module ssrf does not use the custom external_endpoint whether it is set with '--endpoint' or '--external-endpoint'.

The payload is always sent to 'http://wapiti3.ovh'

The custom internal_endpoint is working fine.

bretfourbe:
Fix #63


devl00p:
Nice you spot this one

Issue #65, opened 2021-02-12
Add Fingerprint Web Server to Information Level


fwininger:
When wappalyzer detect a version for a web server like Apache / Nginx, add a WSTG-INFO-02 issue.

https://owasp.org/www-project-web-security-testing-guide/v41/4-Web_Application_Security_Testing/01-Information_Gathering/02-Fingerprint_Web_Server.html

PR #70, opened 2021-02-20
Add vuln for web server with version in wapp module

bretfourbe:
Fix https://github.com/wapiti-scanner/wapiti/issues/65


wapitiCore/definitions/fingerprint_webserver.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+# This file is part of the Wapiti project (https://wapiti.sourceforge.io)
+# Copyright (C) 2021 Nicolas Surribas
```
fwininger, line 5: @bretfourbe change this with our name or remove all the licence part.

wapitiCore/attack/mod_wapp.py
```diff
@@ -90,15 +91,27 @@ def attack(self):
...
+            versions = detected_applications[application_name]["versions"]
+            categories = detected_applications[application_name]["categories"]
+
             self.log_blue(
                 MSG_TECHNO_VERSIONED,
                 application_name,
-                detected_applications[application_name]["versions"]
+                versions
             )
             self.add_addition(
```
devl00p, line 103: This function call should be made inside an `else` block bellow otherwise the web server may appear in both categories.

wapitiCore/attack/mod_wapp.py
```diff
@@ -90,15 +91,27 @@ def attack(self):
...
                 category=TECHNO_DETECTED,
                 level=LOW_LEVEL,
                 request=request,
                 info=json.dumps(detected_applications[application_name])
             )
+
+            if versions and "Web servers" in categories:
```
devl00p, line 110: If think it should only be `if "Web servers" in categories` as we probably want to use `WEB_SERVER_VERSIONED` even if the version is missing (?)

devl00p (changes requested):

wapitiCore/attack/mod_wapp.py
bretfourbe: @fwininger do we need web server detection both in vulnerability and additional parts of the report ?

wapitiCore/attack/mod_wapp.py
bretfourbe: I think it is a security issue only if we can determine the version in use since it is something that you can hide.

devl00p: I get it. Then it would make sense to have it in additional too (with or without version)

PR #71, opened 2021-02-21
Fix version detection issue in wappalyzer

bretfourbe:
Fix issue in wappalyzer due to lazy evaluation for boolean expressions.


fwininger (approved):
the code architecture is not great for having this kind of modification but it should work.

PR #77, opened 2021-02-27
Add boolean-based blind SQL injection

devl00p:
Currently Wapiti use two separate modules for SQL injection :
- sql for error-based sql injections
- blindsql for time based sql injections

Both have pros and cons (error-based is very fast but errors are hidden most of the time so detection rate is low while time-based is slow but works with silent errors). A boolean-based module should bring advantages of both:

- it is not that fast but the amount of payloads is reduced compared to time-based, also we don't have to wait for a timeout :+1: 
- works well for blind injection and is more dbms agnostic than time-based (which is using sleep() for mysql and waitfor for mssql for example while other dbms doesn't have such capability)

Also we can move the old blindsql module to a renamed timesql with will be more explicit and make it optional.

Encoding of values should also use Python's quote() function instead of quote_plus() as it seems the '+' sign in URL produces unexpected errors for ASP webpages (while %20 works well)

### DIFF
--- a/make_exe.py
+++ b/make_exe.py
@@ -97,7 +97,7 @@ if a script is vulnerable.""",
             "includes": [
                 "wapitiCore.attack.mod_backup",
                 "wapitiCore.attack.mod_brute_login_form",
-                "wapitiCore.attack.mod_blindsql",
+                "wapitiCore.attack.mod_timesql",
                 "wapitiCore.attack.mod_buster",
                 "wapitiCore.attack.mod_cookieflags",
                 "wapitiCore.attack.mod_crlf",
--- a/tests/attack/test_mod_sql.py
+++ b/tests/attack/test_mod_sql.py
@@ -200,3 +200,32 @@ def test_blind_detection():
         assert persister.vulnerabilities
         # One request for error-based, one to get normal response, four to test boolean-based attack
         assert len(responses.calls) == 6
+
+
+@responses.activate
+def test_negative_blind():
+    responses.add(
+        responses.GET,
+        url="http://perdu.com/",
+        body="Hello there"
+    )
+
+    persister = FakePersister()
+
+    request = Request("http://perdu.com/?foo=bar")
+    request.path_id = 1
+
+    crawler = Crawler("http://perdu.com/", timeout=1)
+    options = {"timeout": 10, "level": 1}
+    logger = Mock()
+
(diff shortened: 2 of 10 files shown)


devl00p:
example of output

```
[*] Lancement du module sql
[+] GET http://testasp.vulnweb.com/showthread.asp?id=1 (0)
[¨] GET http://testasp.vulnweb.com/showthread.asp?id=1%C2%BF%27%22%28 (0)
---
Une erreur HTTP 500 a été obtenue avec http://testasp.vulnweb.com/showthread.asp
Evil request:
    GET /showthread.asp?id=1%C2%BF%27%22%28 HTTP/1.1
    Host: testasp.vulnweb.com
---
[¨] GET http://testasp.vulnweb.com/showthread.asp?id=1%20AND%2079%3D134%20AND%2016%3D16 (0)
[¨] GET http://testasp.vulnweb.com/showthread.asp?id=1%20AND%2079%3D79%20AND%2016%3D16 (0)
[¨] GET http://testasp.vulnweb.com/showthread.asp?id=1%20AND%2039%3D59%20AND%2086%3D86 (0)
[¨] GET http://testasp.vulnweb.com/showthread.asp?id=1%20AND%2039%3D39%20AND%2086%3D86 (0)
---
Injection SQL dans http://testasp.vulnweb.com/showthread.asp via une injection dans le paramètre id
Evil request:
    GET /showthread.asp?id=1%20AND%2039%3D39%20AND%2086%3D86 HTTP/1.1
    Host: testasp.vulnweb.com
---
```

simonbuehler:
hi,

when running wapiti against a login form it finds:
```
Evil request:
    POST /xx/index.php?content=login&action=%27+or+sleep%287%29%3D%27&view_rights=no HTTP/1.1
    Host: www.xxxxx
    Referer: https://wwwxxxxxxx/
    Content-Type: application/x-www-form-urlencoded

    red_content=login&login=&password=Letm3in_&Submit=Login
```

but sqlmap find it's not vulnerable - what is missing in wapiti to verify a real vulnerability?
greets!

devl00p:
Hello !

You can use the curl command line provided in the generated report to see if you can reproduce the behavior.

But time based blind sql injection may be false positives too.

Also with sqlmap be sure to set `risk` and `level` values to maximum to not miss something

PR #94, opened 2021-03-25
.Adding Unique Constraint(path+method) on Path SQLite DB and Applying Changes accordinly

Yvtq8K3n:
In order to resolve the issue: Session is duplicating scan values.
I added Unique constraints on paths table and applied changes in order to only allow insertion if the path+method is not yet present in one of the table records.

Wapiti in this current version is only capable of running 1 scan. Having the possibility to run multiple scans is quite a necessity.  Sometimes I want to scan without being logged on or sometimes logged with different accounts.


fwininger (changes requested):
Please clean your MR, we dont accept something with your environnement

```
VIRTUAL_ENV="/home/marquez/Desktop/Untitled Folder/wapiti/wapiti3"
```

devl00p:
The uniqueness of a request is not limited to the path and method. There
are also response headers and each entry may be tied to different
parameters.

The issue requires more investigation and I will take a look when I have
some time but I don't think a constraint may help Right now as it would
certainly cause other issues.

Le ven. 26 mars 2021 à 23:42, Florian Wininger ***@***.***> a
écrit :

devl00p:
closed, fixed in sqlite persister

PR #95, opened 2021-03-26
Add js parser for angular

bretfourbe:
Add basic js parser for angular : extract related domains and paths.
Add brotli module in requirements to support br content-encoding, this will help to detect more js files.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -18,6 +18,7 @@ yaswfp = ">=0.9.3"
 six = ">=1.15.0"
 browser-cookie3 = "==0.11.4"
 importlib_metadata = "==3.7.2"
+brotli = ">=1.0.9"
 
 [requires]
 python_version = "3.8"
--- a/wapitiCore/net/jsparser_angular.py
+++ b/wapitiCore/net/jsparser_angular.py
@@ -1,0 +1,96 @@
+import re
+from urllib.parse import urlparse
+from tld import get_fld
+from tld.exceptions import TldDomainNotFound, TldBadUrl
+
+
+class JsParserAngular:
+    """
+    A class that try to parse JS files from angular webapps
+    and extract interesting subdomains and paths.
+    """
+
+    def __init__(self, url: str, data: str):
+        """
+        Create a new JsParserAngular object.
+
+        Takes the following arguments:
+            url: The absolute URL of the JS file on the server
+            data: The content of the JS file as a string
+        """
+        self.links = []
+
+        parts = urlparse(url)
+        netloc = parts.netloc
+        self.target_domain = netloc
+
+        # if netloc and len(netloc.split('.')) > 2:
+        #     netloc = '.'.join(netloc.split('.')[1:])
+
+        # #self.domain = netloc
+        self.domain = self.get_domain(url)
+        self.scheme = parts.scheme
+
+        if self.domain:
+            self.parse_data(data)
+
+    @staticmethod
+    def get_domain(url):
+        domain = None
+        try:
+            domain = get_fld(url)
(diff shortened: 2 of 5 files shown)


wapitiCore/net/jsparser_angular.py
```diff
@@ -0,0 +1,95 @@
...
+        for path in data_found:
+            if path and "http" not in path and path not in path_found:
+                path_found.append(path)
+
+        # re.findall(r"(?:^\w+$|(?:https?:\/)?[\w\.\/%]*(?:\/|%2F|%2f)(?:[\w:?%#~$][\w\-\.@?=%&:/~#]*[\w=/~]))", values)
+
+        # catch full urls
+        # https://www.i2tutorials.com/match-urls-using-regular-expressions-in-python/
+        urls = \
+            re.findall(r"https?:\/\/(?:[\w\-_]+(?:(?:\.[\w\-_]+)+))(?:[\w\-\.,@?^=%&:/~\+#]*[\w\-\@?^=%&/~\+#])?", data)
```
jinxka, line 78: `r"https?:\/\/[^\"'\\ )]+"`

jinxka (changes requested):

Pipfile
```diff
@@ -18,6 +18,7 @@ yaswfp = ">=0.9.3"
 six = ">=1.15.0"
 browser-cookie3 = "==0.11.4"
 importlib_metadata = "==3.7.2"
+brotli = ">=1.0.9"
```
devl00p, line 21: I don't see where you are using this but if the objective is being able to decode brotli encoded HTTP requests it seems we should use `httpx[brotli]` as dependency

see https://www.python-httpx.org/#installation

wapitiCore/net/jsparser_angular.py
```diff
@@ -0,0 +1,81 @@
...
+            url: The absolute URL of the JS file on the server
+            data: The content of the JS file as a string
+        """
+        self.links = []
+
+        parts = urlparse(url)
+        netloc = parts.netloc
+        self.target_domain = netloc
+
+        if netloc and len(netloc.split('.')) > 2:
```
devl00p, line 25: You should not extract domain by splitting on dots, it won't work for .co.uk suffixes. Instead use `tld.get_fld` which is already used elsewhere in Wapiti.

Also you should check that parts.hostname is not an IP address.

wapitiCore/net/jsparser_angular.py
```diff
@@ -0,0 +1,81 @@
...
+            netloc = '.'.join(netloc.split('.')[1:])
+
+        self.domain = netloc
+        self.scheme = parts.scheme
+
+        if self.domain:
```
devl00p, line 31: Is there any possible case where `self.domain` is not set ?

devl00p (changes requested):

Pipfile
bretfourbe: `requests` need `brotli` to decode br encoded responses. Many js files are br encoded, that is why i need it. But now we are using `httpx` so we should use `http[brotli]` as you said.

wapitiCore/net/jsparser_angular.py
bretfourbe: Only if we have an invalid url

Pipfile
bretfourbe: But it appears that `httpx[brotli]` does not work on debian... and `brotli` works

devl00p: Indeed I checked with Ubuntu and openSUSE and `pip install brotli` seems to be enough

devl00p (approved):

PR #102, opened 2021-03-31
Add module to detect Drupal version

OussamaBeng:
Adding module to detect Drupal version by comparing the hashed files (.js , .css and .txt) of different versions


wapitiCore/attack/mod_drupal_enum.py
```diff
@@ -0,0 +1,117 @@
...
+
+            cont = rep.content.encode()
+            hash_content = hashlib.sha256(cont).hexdigest()
+
+            if uri in data :
+                if hash_content in data[uri] :
+                    vers[uri] = data[uri][hash_content]
+
+        i = 0
+        result =[]
```
devl00p, line 47: you can do
`self.versions = set.intersection(*[set(versions) for versions in vers.values()])`

wapitiCore/attack/mod_drupal_enum.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def attack(self, request: Request):
+        self.finished = True
+        request_to_root = Request(request.url)
+        response = self.crawler.send(request_to_root, follow_redirects=True)
+
+        if self.check_drupal(request_to_root.url):
+            self.detect_version(request_to_root.url)
+            if self.versions :
+                self.versions = sorted(self.versions, key=lambda x: [i for i in x.split('.')])
```
devl00p, line 82: You may be interested in the Python `natsort` module but if should be enough here

OussamaBeng: I tried to do it with natsort  but i had an import problem

Issue #115, opened 2021-04-25
Add a build of the Dockerfile in github action


fwininger:
The master branch are not able to build the projet Dockerfile with de brower-cookie3 dependency. 

Please add a build of the Dockerfile in github action.

devl00p:
It seems related to https://github.com/pyca/cryptography/issues/5753

I managed to get the image working with the following dockerfile (just ading `python3-pip` to packages to install using apt):

```dockerfile
FROM debian:buster-slim as build

ENV DEBIAN_FRONTEND=noninteractive \
  LANG=en_US.UTF-8

WORKDIR /usr/src/app

RUN apt update
RUN apt install python3 python3-setuptools python3-pip ca-certificates -y \
  && apt clean -yq \
  && apt autoremove -yq \
  && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* \
  && truncate -s 0 /var/log/*log

COPY . .

RUN python3 setup.py install

FROM debian:buster-slim

ENV DEBIAN_FRONTEND=noninteractive \
  LANG=en_US.UTF-8 \
  PYTHONDONTWRITEBYTECODE=1

RUN apt update
RUN apt install python3 python3-setuptools python3-pip -y \
  && apt clean -yq \
  && apt autoremove -yq \
  && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* \
  && truncate -s 0 /var/log/*log

COPY --from=build /usr/local/lib/python3.7/dist-packages/ /usr/local/lib/python3.7/dist-packages/
COPY --from=build /usr/local/bin/wapiti /usr/local/bin/wapiti-getcookie /usr/local/bin/

CMD ["wapiti"]
```

fwininger:
close with #120

PR #116, opened 2021-04-25
WIP: Dockerfile: Fix `setup install` problem with build of cryptography (required by browser_cookie3)

devl00p:


fwininger:
Thanks for fix the issue, but we don't want to include python-pip depencendy.

For our project, we don't care about the browser-cookie extention and don't want to include 349MB of python binaire for nothing.

```
REPOSITORY                                                       TAG           IMAGE ID       CREATED        SIZE
after                                                            latest        f8933a6a33be   2 hours ago    489MB
before                                                           latest        4fda28c8dd47   2 hours ago    140MB
```

Our current solution is a quick and dirty patch to fix the version of cryptography to 3.3.2. "cryptography==3.3.2"

fwininger:
close with #115

Issue #109, opened 2021-04-09
Add necessary details in output result to retest a specific issue


bretfourbe:
Output contains http request and curl command but there is not enough information to retest a specific issue with wapiti without crawling again. 
![image](https://user-images.githubusercontent.com/27951656/114158235-c2bd7280-9924-11eb-9715-2d336bb9cb91.png)

For instance, it might be interesting to have the module name, the referer and maybe the credentials needed to reproduce the issue via wapiti.

devl00p:
Indeed module and referer may be useful but I'm not sure about the credentials, they are normally used everywhere for the requests and given using command line so the user knows them. Also keeping some credentials somewhere may be dangerous.

I guess it would be more useful to include cookies to reproduce the issue. What do you think?

bretfourbe:
The aim is to help in automatically relaunching a wapiti scan with specific parameters.
Cookies might not be valid for a long time, but yes keeping credentials in reports may be dangerous i think.

devl00p:
To me it was more in the sense "being able to reproduce the request" hence giving cookies (of course that may expired though).

I would prefer if the credentials doesn't appear. We may put a placeholder somewhere if we want though (like `-u USER:PASS` for the Curl PoC). We may also include proxy options this way (I don't remember if it is done already)

PR #124, opened 2021-06-11
Refactor fake persister and add information in output file to help retesting an issue with wapiti

bretfourbe:
Refactor fake persister in tests + add add_payload methods in attack.py and update attack modules to support these changes.
Add module, referer and auth information in output file to help restesting a specific issue with wapiti.
Rename cookieflags test module file, refactor csp, drupal_enum and http_headers attack modules.
Fix some pylint errors.

Fix https://github.com/wapiti-scanner/wapiti/issues/109


devl00p:
@bretfourbe can we synchronize on this issue ? Unfortunately I was working exactly on the same components for another MR ( https://github.com/wapiti-scanner/wapiti/pull/126 ) and I completely removed the `FakePersister`.

I can accept your MR but most of it will be rewritten with mine next. Maybe you can sum up the changes in yours so I make sure I have those in my MR too.

wapitiCore/attack/mod_ssrf.py
```diff
@@ -58,7 +58,7 @@ def mutate(self, request: Request):
         # self._attacks_per_url_pattern[request.hash_params] += estimation
 
         for params_list in [get_params, post_params, file_params]:
-            for i in range(len(params_list)):
+            for i, _ in enumerate(params_list):
```
devl00p, line 61: You should use double underscore otherwise you are overwriting the translation function

wapitiCore/attack/attack.py
```diff
@@ -230,6 +232,51 @@ def set_verbose(self, verbose):
     def set_color(self):
         self.color = 1
 
+    def add_payload(self, request_id: int, payload_type: str, category: str,
+    level=0, request=None, parameter="", info=""):
+        self.persister.add_payload(request_id, payload_type, self.name, category, level, request, parameter, info)
+
+    def add_payload_critical(self, request_id: int, payload_type: str, category: str,
```
devl00p, line 239: You can use [functools.partialmethod](https://docs.python.org/3/library/functools.html#functools.partialmethod) for all those methods :)

PR #126, opened 2021-06-15
Persister: Use SQLAlchemy in order to support other DBMS / asyncio mode / Remove FakePersister

devl00p:
First:

In order to support other databases than SQLite for the persister we should use SQL Alchemy that offers a great level of abstraction.

SQLAlchemy ORM seems clumsy to use for our situation, especially as we need to create Request objects without necessarily saving them to a database. Also the ORM generates some queries to the database when we do not expect them.

SQLAlchemy Core seems the good solution: we can specify abtract column types and the library will use the good data types for each DBMS.

Second:

SQLAlchemy offers a asyncio interface so we can use the dbms in a non blocking way and prevent blocking the asyncio event loop. There is a bit of magic for sqlite because asyncio is made for network and sqlite is file based but with the aiosqlite engine queries are executed inside threads.

Third:

The FakePersister used in tests in non-pythonic and the code appears at too many places. As I had to rewrote this because of asyncio I prefered to completely remove it in favor of standard Python mocking.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -19,6 +19,9 @@ six = ">=1.15.0"
 browser-cookie3 = "==0.11.4"
 importlib_metadata = "==3.7.2"
 brotli = ">=1.0.9"
+sqlalchemy = ">=1.4.17"
+aiocache = "==0.11.1"
+aiosqlite = "==0.17.0"
 
 [requires]
 python_version = "3.8"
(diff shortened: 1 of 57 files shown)


devl00p:
Also: improvements on sql requests to make saving scan state faster as well as marking http resources attacked.

devl00p:
Also note that this is the first part of supporting multiple DBs but it should be the biggest by far.

Next steps:

- allows to set a prefix for table names
- load dbms configuration from a config file (ini ? toml ?)
- test the whole stuff for postgres and mysql with async engines

PR #130, opened 2021-06-25
Add module http_post

OussamaBeng:
Check if credentials are transported over an encrypted channel


wapitiCore/attack/mod_http_post.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+# This module check the security of transported credentials of login forms
+class mod_http_post(Attack):
+    """Check if credentials are transported on an encrypted channel."""
+    name = "http_post"
+
+    async def must_attack(self, request: Request):
+        # We leverage the fact that the crawler will fill password entries with a known placeholder
+        if "Letm3in_" not in request.encoded_data + request.encoded_params:
```
devl00p, line 16: I think you can directly discard on the protocol here too

wapitiCore/attack/mod_http_post.py
```diff
@@ -0,0 +1,46 @@
...
+            return False
+
+        # We may want to remove this but if not available fallback to target URL
+        if not request.referer:
```
devl00p, line 20: I don't think you need to check the referer, protocol and "Letm3in_" should be enough

wapitiCore/attack/mod_http_post.py
```diff
@@ -0,0 +1,46 @@
...
+            return False
+
+        return True
+
+    async def attack(self, request: Request):
+        try:
+            page = await self.crawler.async_send(request, follow_redirects=True)
```
devl00p, line 27: I don't see any benefit into sending the request again.
Especially the form may be at another URL than the target of the form.

devl00p: You will make things easier and more effective just by checking requests with clear protocol and presence of "Letm3in_".

devl00p (changes requested):

wapitiCore/attack/mod_http_post.py
```diff
@@ -0,0 +1,36 @@
...
+        if "https://" in request.url:
+            return False
+
+        return True
+
+    async def attack(self, request: Request):
+        
+        if "Letm3in_" not in request.encoded_data + request.encoded_params:
+            return
+        self.finished = True
```
devl00p, line 25: I think you can remove this line

devl00p: Also as an improvement we can keep track of files we marked as vulnerable (using `request.path`) to prevent duplicates (you can try on http://www.tvsoop.com/ for example which will generate lot of duplicates)

wapitiCore/attack/mod_http_post.py
```diff
@@ -0,0 +1,36 @@
...
+
+        self.log_red(NAME)
```
devl00p, line 27: Please use the output format used by other modules (take example here https://github.com/wapiti-scanner/wapiti/blob/master/wapitiCore/attack/mod_ssrf.py#L253 )

wapitiCore/definitions/http_post.py
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+from wapitiCore.language.language import _
+
+TYPE = "vulnerability"
+NAME = _("POST HTTP")
```
devl00p, line 7: Missing translations in en / fr .po files

devl00p (changes requested):

bretfourbe:
I think this one can be closed now, replaced by module `https_redirect` : https://github.com/wapiti-scanner/wapiti/pull/411 ?

devl00p:
we now have a passive module for that

PR #131, opened 2021-06-25
Add log option to generate logfile and replace custom logger by logger from loguru

bretfourbe:
Add log option to generate logfile and replace custom logger by logger from loguru.
Update test files to support these changes.


devl00p:
Nice, thank you

PR #144, opened 2021-08-19
Subdomain takeover module

### LINKED ISSUE #143: Subdomain takeovers module
I made a PoC called [tekover](https://github.com/devl00p/tekover) using asyncio.

Writing a Wapiti module for this should be (almost) as easy as ctrl+c / ctrl+v

devl00p:
This module brings subdomain takeover vulnerability detection.

It relies on some files (wordlist of subdomains, list of DNS resolvers, json file of rules to deal with special cases of takeovers)

Fix #143

Tests for 3.7 do not work on github CI, can't tell why yet


devl00p:
> Tests for 3.7 do not work on github CI, can't tell why yet

I did some tests with Python 3.7 installed with Pyenv and it turns out some of the mocks do not work with this version of Python.

I guess it is related to asyncio. Patching the objects does not have any effect, the real classes / methods are called instead.

devl00p:
> I guess it is related to asyncio. Patching the objects does not have any effect, the real classes / methods are called instead.

So it was kinda related to this behavior: https://github.com/pytest-dev/pytest-asyncio/issues/42

Looks like `@pytest.mark.asyncio` doesn't like to share its space with `@patch` decorators.

So dns queries weren't mocked and another strange effect was that the `respx.route()` call was unleashed and captured requests of all remaining tests :laughing: 

Using `patch()` as a context (not as a decorator) was enough to fix the issue.

bretfourbe (commented):
Interesting ! But do we need such big files for resolvers.txt and takeover_finerprints.json since the module only treats github and shopify services at the moment ?

devl00p:
> Interesting ! But do we need such big files for resolvers.txt and takeover_finerprints.json since the module only treats github and shopify services at the moment ?

Other services from `takeover_finerprints.json` are checked too, see the `return True` at line 149.

But I noticed the false positives cases for github and myshopify hence the additional checks.

For `resolvers.txt` indeed it is big but necessary as most resolvers will block you if you ask them too many questions :sweat_smile: That's why the load is balanced on all resolvers.

PR #153, opened 2021-08-28
Add test for net/page

polyedre:
This adds a test for the make_absolute function of file page.py before a possible refactor:

```
$ pylint wapitiCore/net/page.py
************* Module wapitiCore.net.page
wapitiCore/net/page.py:355:4: R0911: Too many return statements (9/6) (too-many-return-statements)
wapitiCore/net/page.py:721:4: R0912: Too many branches (41/40) (too-many-branches)
```

I noticed that some results are strange to me:

Shouldn't:
 - `'http://base.url/with_folder' + '.'` be `'http://base.url/with_folder'` instead of `http://base.url/` ?
 - `'http://base.url' + '/..//'` be `''` instead of `'http://base.url/'` ?
 - `'http://base.url' + './wrong_folder/../good_folder/'` be `'http://base.url/'` instead of `'http://base.url/good_folder/'` ?

Maybe the function [urllib.parse.urljoin](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin) can be used instead ?

### DIFF
--- a/tests/net/test_page.py
+++ b/tests/net/test_page.py
@@ -1,0 +1,33 @@
+#!/usr/bin/env python3
+
+from httpx import Response, Request
+from wapitiCore.net.page import Page
+
+def test_make_absolute():
+
+    TEST_CASES = [
+        ("http://base.url", "relative", "http://base.url/relative"),
+        ("http://base.url", ".", "http://base.url/"),
+        ("http://base.url/with_folder", ".", "http://base.url/"),
+        ("http://base.url/with_folder", "./with_dot", "http://base.url/with_dot"),
+        ("http://base.url/with_folder", "..", "http://base.url/"),
+        ("http://base.url/with_folder", "../folder", "http://base.url/folder"),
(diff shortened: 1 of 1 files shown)


devl00p:
> Shouldn't:
> 
> * `'http://base.url/with_folder' + '.'` be `'http://base.url/with_folder'` instead of `http://base.url/` ?

As the `with_folder` doesn't have a trailing slash it is considered a file hence the result.

> * `'http://base.url' + '/..//'` be `''` instead of `'http://base.url/'` ?
> * `'http://base.url' + './wrong_folder/../good_folder/'` be `'http://base.url/'` instead of `'http://base.url/good_folder/'` ?

Those are behaviors applied by browsers, I copied the way they work.

> Maybe the function [urllib.parse.urljoin](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin) can be used instead ?

Indeed. When I made this method some years ago I didn't knew about `urljoin` (not sure it existed at that time). Having those tests will help to make sure there is no regression if we move to urljoin

PR #159, opened 2021-09-13
More control over Wapiti generated output with its method

### LINKED ISSUE #156: Loguru doesn't work as expected
I bumped into that behavior using my own mass scanning scripts.

I made this short example to show how to reproduce:

```python
from urllib.parse import urlparse                                                                                      
import sys                                                                                                             
import asyncio                                                                                                         
                                                                                                                       
from wapitiCore.main.wapiti import Wapiti, logging                                                                     
                                                                                                                       
                                                                                                                       
async def scan(url: str):                                                                                              
    wapiti = Wapiti(url)                                                                                               
    wapiti.set_max_scan_time(120)                                                                                      
                                                                                                                       
    wapiti.verbosity(2)                                                                                                
    wapiti.set_timeout(20)                                                                                             
    wapiti.set_modules("xss")                                                                                          
    wapiti.set_bug_reporting(False)                                                                                    
                                                                                                                       
    parts = urlparse(url)                                                                                              
    wapiti.set_output_file(f"/tmp/{parts.scheme}_{parts.netloc}.json")                                                 
    wapiti.set_report_generator_type("json")                                                                           
                                                                                                                       
    wapiti.set_attack_options({"timeout": 20, "level": 1})                                                             
                                                                                                                       
    stop_event = asyncio.Event()                                                                                       
    await wapiti.init_persister()                                                                                      
    await wapiti.flush_session()                                                                                       
    await wapiti.browse(stop_event, parallelism=64)                                                                    
    await wapiti.attack(stop_event)                                                                                    
                                                                                                                       
if __name__ == "__main__":                                                                                             
    logging.info("Info")                                                                                               
    logging.warning("Warning")                                                                                         
    logging.error("Error")                                                                                             
    logging.critical("Critical")                                                                                       
    asyncio.run(scan(sys.argv[1]))                                                                                     
```

As the verbosity is set to 2 we should see URLs as they are crawled but instead nothing happens (only "Report" is printed because it uses the builtin `print()` function).

I debugged the code in `crawler.py` at this point:

```python
            if self._log:
                logging.info("[+] {0}".format(request))
```

`self._log` is `True` and `logging.info` is called but nothing happens on the console.

The same happens in the `__main__` of the script with no output generated for all logging calls.

I tested on two computers to be sure it is not related to my environment.

Now I can't figure out how Wapiti does manage to generate some output as I took the same order of calls for methods of the `Wapiti` class.

Maybe @bretfourbe you will have an idea ?

Also I played with [Rich](https://github.com/willmcgugan/rich#readme) recently and it has more features :yum:

devl00p:
Fix #156 

I moved log_red and co from attack.py to log.py directly.

I added a refresh_logging method that reconfigure logging and is called when verbosity / coloring / log file is changed.

I removed the need to pass verbosity and color to attack modules as they are now handled directly by logging level.

### DIFF
--- a/wapitiCore/main/log.py
+++ b/wapitiCore/main/log.py
@@ -22,7 +22,13 @@ from functools import partial
 from loguru import logger as logging
 
 logging.remove()
-# logging.debug is 10
+
+# Setup additional logging levels, from the less important to the more critical
+# Each attempted mutated request will be logged as VERBOSE as it generates a lot of output
+# Each attacked original request will be logged as INFO
+# Others info like currently used attack module must be logged even in quiet mode so BLUE level must be used as least
+
+# logging.debug is level 10, this is the value defined in Python's logging module and is reused by loguru
 logging.level("VERBOSE", no=15)
 # logging.info is 20
 logging.level("BLUE", no=21, color="<blue>")
--- a/wapitiCore/main/log.py
+++ b/wapitiCore/main/log.py
@@ -16,11 +16,29 @@
 # You should have received a copy of the GNU General Public License
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+import sys
+from functools import partial
 
 from loguru import logger as logging
 
 logging.remove()
-logging.level("RED", no=45, color="<red>")
-logging.level("ORANGE", no=35, color="<yellow>")
-logging.level("GREEN", no=22, color="<green>")
+# logging.debug is 10
+logging.level("VERBOSE", no=15)
+# logging.info is 20
 logging.level("BLUE", no=21, color="<blue>")
+logging.level("GREEN", no=22, color="<green>")
+# logging.success is 25
+# logging.warning is 30
+logging.level("ORANGE", no=35, color="<yellow>")
+# logging.error is 40
+logging.level("RED", no=45, color="<red>")
+# logging.critical is 50
+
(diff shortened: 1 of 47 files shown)


wapitiCore/main/wapiti.py
```diff
@@ -123,7 +123,7 @@ def __init__(self, root_url, scope="folder", session_dir=None, config_dir=None):
         self.forms = []
         self.attacks = []
 
-        self.color = 0
+        self.color = False
```
polyedre, line 126: Is it a boolean that represents whether colors are supported by the terminal? If so, would it be possible to rename the variable to something more explicit like `self.color_enabled` or something?

wapitiCore/main/log.py
```diff
@@ -16,11 +16,29 @@
...
+logging.level("VERBOSE", no=15)
+# logging.info is 20
 logging.level("BLUE", no=21, color="<blue>")
+logging.level("GREEN", no=22, color="<green>")
+# logging.success is 25
+# logging.warning is 30
+logging.level("ORANGE", no=35, color="<yellow>")
+# logging.error is 40
+logging.level("RED", no=45, color="<red>")
+# logging.critical is 50
```
polyedre, line 35: Maybe we could add a comment here to explain what are these numbers and why are they set to these values?

polyedre (approved):
Removing the verbosity management from the Attack class really seems to be an improvement ! lgtm

bretfourbe (approved):

PR #163, opened 2021-09-27
mod_sql: change order of boolean tests to prevent false-positives

### LINKED ISSUE #158: Faux positif lors de codes de retour 303 sur les injections SQL
Lors de tentatives d'injections SQL, un code de retour 303 est pris comme une injection SQL réussie.

Cela provoque des faux positifs, notamment sur les outils Moodle, où le formulaire d'authentification répond 303 à des requêtes d'injections.

devl00p:
mod_sql: change order of boolean tests to prevent false-positives (Moodle case)

Fix #158

### DIFF
--- a/tests/attack/test_mod_sql.py
+++ b/tests/attack/test_mod_sql.py
@@ -240,6 +240,12 @@ async def test_blind_detection_parenthesis():
         await module.attack(request)
 
         assert persister.add_payload.call_count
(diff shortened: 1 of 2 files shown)


devl00p:
Use two tests which should give no results (1 = 2) then two tests that should give results (1 = 1) instead of False / True / False / True like previously

bretfourbe (approved):

PR #166, opened 2021-09-29
Fix Nikto false positives that are relying only on status code

### LINKED ISSUE #139: False positives in modules nikto and buster
Some servers are configured to reply with a 200 response even if a file is missing.

Such false positives should be checked and fixed in both modules. For example by trying to query an unlikely existing document and comparing the response.

Example of line in nikto_db that may cause such behavior:

`001924,3092,1,/weblog/,GET,200,,,,,This might be interesting.,,`

devl00p:
Fix #139

### DIFF
--- a/wapitiCore/attack/mod_nikto.py
+++ b/wapitiCore/attack/mod_nikto.py
@@ -22,11 +22,12 @@ import re
 import os
 import random
 from urllib.parse import urlparse
+from typing import List
 
 from httpx import RequestError
 
 from wapitiCore.main.log import logging, log_verbose, log_red
-from wapitiCore.attack.attack import Attack
+from wapitiCore.attack.attack import Attack, random_string
 from wapitiCore.language.vulnerability import _
 from wapitiCore.definitions.dangerous_resource import NAME
 from wapitiCore.net.web import Request
@@ -79,6 +80,9 @@ class mod_nikto(Attack):
         if not os.path.isdir(self.user_config_dir):
             os.makedirs(self.user_config_dir)
 
+        self.status_codes = {}
+        self.random_string = random_string()
+
     async def update(self):
         """Update the Nikto database from the web and load the patterns."""
         try:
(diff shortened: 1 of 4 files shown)


devl00p:
I tested Nikto (the original tool) and saw that it was filtering false positives so the issue was indeed on our side, not in the Nikto DB.

At start Nikto makes several requests to see the webserver behavior:

```
127.0.0.1 - - [29/Sep/2021 22:11:18] "GET /snSWx2fQ.exe HTTP/1.1" 200 -
127.0.0.1 - - [29/Sep/2021 22:11:18] "GET /snSWx2fQ HTTP/1.1" 200 -
127.0.0.1 - - [29/Sep/2021 22:11:18] "GET /snSWx2fQ.TXT HTTP/1.1" 200 -
127.0.0.1 - - [29/Sep/2021 22:11:18] "GET /.snSWx2fQ HTTP/1.1" 200 -
127.0.0.1 - - [29/Sep/2021 22:11:18] "GET /snSWx2fQ/ HTTP/1.1" 200 -
```

The obtained status codes are used to check for false positives later.

While it may seems inaccurate to only test those files at the root of the webserver and not the directory of every entry in the `nikto_db` file, it actually saves a LOT of unecessary requests that's why I chose to keep this behavior.

I improved the verifications slightly because instead of using a list of hardcoded suffixes (exe, txt) `mod_nikto` will test for an improbable file with the same suffix than the payload.

I noticed that `mod_buster` already had a false positive check so either I forgot it or I did not record a special case I bumped into but at least I added 429 status codes (for `too many requests`) to the ignore list.

wapitiCore/net/web.py
```diff
@@ -479,6 +479,10 @@ def file_path(self):
     def is_root(self) -> bool:
         return self._file_path == "/"
 
+    @property
+    def root(self) -> str:
```
bretfourbe, line 483: Maybe use something less ambigous than `root` ?

wapitiCore/attack/mod_nikto.py
```diff
@@ -105,6 +109,26 @@ async def must_attack(self, request: Request):
 
         return request.url == await self.persister.get_root_url()
 
+    async def is_false_positive(self, evil_request: Request, expected_status_codes: List[int]) -> bool:
+        # Check for false positives by asking an improbable file of the same type at the root of the server
+        # Use a dict to cache requests
+        if evil_request.is_directory:
+            request = Request(f"{evil_request.root}{self.random_string}/")
+        else:
+            request = Request(f"{evil_request.root}{self.random_string}{evil_request.file_ext}")
```
bretfourbe, line 118: What about `evil_request.dir_name` instead of `evil_request.root` ?

devl00p: I thought about that unfortunately it generates a lot of additional requests so it comes with a big performance penalty.

I saw Nikto does apply checks on the root so I kept this behavior.

wapitiCore/net/web.py
devl00p: `website` ? `webroot` ? :thinking:

bretfourbe: Oh i think i misunderstood the behavior of urljoin with `/`

PR #170, opened 2021-10-06
Fix lint: line-too-long & redefined-argument-from-local & unspecified-encoding

JulienTD:
Hello, in this pull request I have fixed three linting errors:
- line-too-long
- redefined-argument-from-local
- unspecified-encoding

### DIFF
--- a/wapitiCore/attack/attack.py
+++ b/wapitiCore/attack/attack.py
@@ -558,7 +558,7 @@ class PayloadReader:
         """returns a array"""
         lines = []
         try:
-            with open(filename, errors="ignore") as file:
+            with open(filename, errors="ignore", encoding='utf-8') as file:
                 for line in file:
                     clean_line, flags = self.process_line(line)
                     if clean_line:
--- a/wapitiCore/main/wapiti.py
+++ b/wapitiCore/main/wapiti.py
@@ -469,16 +469,9 @@ class Wapiti:
                         await attack_module.attack(original_request)
 
                     if (datetime.utcnow() - start).total_seconds() > self._max_attack_time >= 1:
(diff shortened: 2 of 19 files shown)


wapitiCore/attack/attack.py
```diff
@@ -558,7 +558,7 @@ def read_payloads(self, filename):
         """returns a array"""
         lines = []
         try:
-            with open(filename, errors="ignore") as file:
+            with open(filename, errors="ignore", encoding='utf-8') as file:
```
devl00p, line 561: Wow! I was pretty sure the default value on `open()` for encoding was `utf-8`, thank you for fixing this.

polyedre (approved):
Looks good to me

wapitiCore/main/wapiti.py
```diff
@@ -469,7 +469,12 @@ async def attack(self, stop_event: asyncio.Event):
                         await attack_module.attack(original_request)
 
                     if (datetime.utcnow() - start).total_seconds() > self._max_attack_time >= 1:
-                        logging.info(_("Max attack time was reached for module {0}, stopping.".format(attack_module.name)))
+                        # FIXME: Right now we cannot remove the pylint: disable line because the current I18N system
+                        # uses the string as a token so we cannot use f string
+                        # pylint: disable=consider-using-f-string
+                        logging.info(
```
devl00p, line 475: Won't work here as the formating is taking place before the translation but you can do something like:

```python
                        logging.info(
                            _("Max attack time was reached for module %s, stopping."), attack_module.name
                        )
```

It would require to fix the .po files but it is ok.

devl00p (changes requested):

PR #186, opened 2021-10-25
Configurable DB backend with a config file

devl00p:
Already tested with mysql+asyncmy

This MR is about using an alternative DB to sqlite3 for the persister, not about sql attack modules

### DIFF
--- a/tests/web/test_postgres_backend.py
+++ b/tests/web/test_postgres_backend.py
@@ -240,6 +240,7 @@ async def test_raw_post():
     assert payload.parameter == "z"
 
 
+@pytest.mark.xfail(reason="PostgreSQL can't encode NULL byte to UTF-8")
 @pytest.mark.asyncio
 async def test_null_byte_in_parameter():
     persister = SqlPersister("postgresql+asyncpg://postgres:wapiti@127.0.0.1:5432/wapiti",  "test5_")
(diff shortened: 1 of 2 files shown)


devl00p:
Added services mysql and postgres for integration tests but python can't resolve container names in Github CI :thinking: 

Any idea @bretfourbe @fwininger ?

devl00p:
> Added services mysql and postgres for integration tests but python can't resolve container names in Github CI thinking
> 
> Any idea @bretfourbe @fwininger ?

Nevermind, I solved the issue. It was a runner vs container problem.

devl00p:
Ready for review. The feature to be able to use another backend than sqlite3 is complete even if there are specific issues that may appear with mysql and postgresql. At least it should not have any impact on the default sqlite3 backend.

The feature won't be documented yet and depencies won't be added to the setup right now (even with `extras`). The support of mysql and postgresql but be considered as experimental.

Tests have been added and the existing `test_persister` as been renamed to `test_sqlite_backend`.

I had to put cython in the build dependencies as pip didn't get the wheel for `asyncmy` for unknown reasons and adding it to the setup failed too.

One of the tests for posgresql was marked as XFAIL. Unfortunately fixing it would require lot of work that is not in the scope of this PR:

In order to be able to insert all kind of encoded data in the databases it would require to change the types of parameters names and values to bytes instead of strings. It would then require to directly give bytes (instead of tuples or dict) to httpx then it would mean we would have to parse the URLs and forms in a different way that actually (and Python is not able to do this directly, see https://stackoverflow.com/a/63964650 ). It may be worth to change it later.

PR #191, opened 2021-11-17
Upgrade httpx and httpx-socks versions

### LINKED ISSUE #189: ERROR: httpx 0.20.0 has requirement httpcore<0.14.0,>=0.13.3, but you'll have httpcore 0.14.1 which is incompatible.
When using pip to install wapiti, I get this error when trying to install.

I can see that the `httpcore` library was updated to `0.14.1` recently and as such the `httpx` library was updated to `0.21.0` to address this.

Please can you update the version pin of `httpx` that wapiti uses, so we can get past the installation error, or pin the version of `httpcore` to at least `0.14.0`?

I see these errors in the install log:

```
ERROR: httpx 0.20.0 has requirement httpcore<0.14.0,>=0.13.3, but you'll have httpcore 0.14.1 which is incompatible.
```
This may be because we allow the following ` httpcore>=0.13.7` so will always pick latest:

```
Collecting httpcore>=0.13.7
  Downloading httpcore-0.14.1-py3-none-any.whl (55 kB)
```

httpx release: https://github.com/encode/httpx/releases/tag/0.21.0

devl00p:
Fix #189 

SOCKS support seems better

### DIFF
--- a/wapitiCore/main/wapiti.py
+++ b/wapitiCore/main/wapiti.py
@@ -504,7 +504,7 @@ class Wapiti:
                             logging.info(f"{WAPITI_VERSION}. httpx {httpx.__version__}. OS {sys.platform}")
 
                         try:
-                            with open(traceback_file, "rb", encoding='utf-8') as traceback_byte_fd:
+                            with open(traceback_file, "rb") as traceback_byte_fd:
                                 upload_request = Request(
                                     "https://wapiti3.ovh/upload.php",
                                     file_params=[
--- a/Pipfile
+++ b/Pipfile
@@ -5,16 +5,16 @@ verify_ssl = true
 
 [dev-packages]
 pytest = ">=6.2.2"
-respx = "==0.18.0"
+respx = "==0.19.0"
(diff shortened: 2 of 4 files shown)


wapitiCore/main/wapiti.py
```diff
@@ -504,7 +504,7 @@ async def attack(self, stop_event: asyncio.Event):
                             logging.info(f"{WAPITI_VERSION}. httpx {httpx.__version__}. OS {sys.platform}")
 
                         try:
-                            with open(traceback_file, "rb", encoding='utf-8') as traceback_byte_fd:
+                            with open(traceback_file, "rb") as traceback_byte_fd:
```
bretfourbe, line 507: Why did you remove `encoding` here ?

devl00p: it is not required as we are reading as binary

devl00p: It is not required as we are opening the file in binary mode

bretfourbe (approved):

Issue #157, opened 2021-09-04
Is there a passive mode?


peter17:
Hi,

I have a question that could be in the FAQ or in the main README: is there a "passive" / "non-intrusive" mode that would perform only checks that would not harm the target?

For instance: search for potentially dangerous files on the server, folder and file enumeration, checking HTTP security headers, etc.

Thanks in advance

bretfourbe:
Hi @peter17,

For now, there is no such "passive" mode. You need to use `-m` option with the modules that you want to use to do something similar.

@devl00p do you think we could define a "passive" mode like it is done for "common" for instance ?

devl00p:
Yes indeed, a new group could be added for modules that doesn't use attack payloads.

devl00p:
`-m passive` is available in latest stable release

PR #194, opened 2021-11-24
Added a passive mode to only execute nonharmful modules

JulienTD:
## Description

Added a passive mode with these modules:
- cookieflags
- csp
- http_headers
- wapp
  
It is launched by writing `passive` as an argument to the module list.  
Example: `wapiti -u http://foobar/ -m passive`  
## Related Issue(s)
 
#157 Passive mode

### DIFF
--- a/wapitiCore/attack/attack.py
+++ b/wapitiCore/attack/attack.py
@@ -76,6 +76,14 @@ commons = [
     "xss"
 ]
 
+# Modules that will be used in passive mode -m passive
+passives = [
+    "cookieflags",
+    "csp",
+    "http_headers",
+    "wapp"
+]
+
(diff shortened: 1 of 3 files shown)


wapitiCore/attack/attack.py
```diff
@@ -76,6 +76,14 @@
     "xss"
 ]
 
+# Modules that will be used in passive mode --passive
```
bretfourbe, line 79: `-m passive` not `--passive`

devl00p:
Looks good to me. Maybe add a little test in `tests/cli/test_options.py::test_options`

bretfourbe (approved):

PR #177, opened 2021-10-18
Update auth report

JulienTD:
## Description

This pull request fixes #162 by:
- moving auth info to the `infos` section (previously was written on every request)
- adding two new fields: `logged_in`  that permits to verify if the authentication has been successful (works with post, basic, digest, and NTLM) & `form` that permits to add the two used fields names to connect with `post` connection method.
- adding auth section to all report types (txt, html, csv, json, and xml)

## Related Issue(s)
 
JSON reporting of authentication failures #162

PR #200, opened 2021-11-29
Added a missing error code in the auth method to verify if the user is logged in

JulienTD:
## Description

Added the 403 error in the auth method to verify if the user is logged in

## Related Issue(s)
 https://github.com/wapiti-scanner/wapiti/pull/177#discussion_r756786625


JulienTD:
@devl00p is it possible to relaunch the CI because with the PR (#199) fixed it :rocket:

wapitiCore/net/crawler.py
```diff
@@ -370,7 +370,7 @@ async def async_try_login(self, auth_url: str, auth_type: str) -> Tuple[bool, di
     async def _async_try_login_basic_digest_ntlm(self, auth_url: str) -> Tuple[bool, dict]:
         page = await self.async_get(web.Request(auth_url))
 
-        if page.status in (401, 404):
+        if page.status in (401, 403, 404):
```
polyedre, line 373: I'm not sure if it is relevant to keep the 404 error code here. You can be logged in and receive some 404.

JulienTD: I think it is relevant to keep the code 404, because the method `_async_try_login_basic_digest_ntlm` is used by three types of auth:
- basic
- digest
- ntlm
  
Sometimes, the basic auth can be hidden so it can return either 200 or either 404 (https://httpbin.org/#/Auth/get_hidden_basic_auth__user___passwd_).

devl00p: Seems good to me. Rebase please

JulienTD: done

PR #201, opened 2021-11-29
Added the number of crawled pages in the info section of the report

JulienTD:
## Description

In this pull request, I have added the number of crawled pages in the info section of the report.

## Related Issue(s)
 
#197


devl00p:
OK for me. The information may be put elsewhere in the HTML report but I guess the HTML template would need a more in-depth rework.

PR #202, opened 2021-11-29
Improved the post auth system by excluding automatically the disconnect url

JulienTD:
## Description

In this pull request, the post auth system has been improved by:
- adding the keyword `signoff` to the disconnect detection regex
- by excluding automatically the disconnect url

## Related Issue(s)
 
#198

### DIFF
--- a/wapitiCore/net/crawler.py
+++ b/wapitiCore/net/crawler.py
@@ -74,6 +74,7 @@ EXCLUDED_MEDIA_EXTENSIONS = (
 
 BAD_URL_REGEX = re.compile(r"https?:/[^/]+")
 
+DISCONNECT_REGEX = r'(?i)((log|sign)\s?(out|off)|disconnect|déconnexion)'
 
 def wildcard_translate(pattern):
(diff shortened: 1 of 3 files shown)


bretfourbe (approved):

wapitiCore/net/crawler.py
```diff
@@ -372,13 +376,28 @@ async def _async_try_login_basic_digest_ntlm(self, auth_url: str) -> Tuple[bool,
...
-        return True, {}
+        return True, {}, []
+
+    @staticmethod
+    def _extract_disconnect_urls(page: Page) -> List[str]:
+        """
+        Extract all the disconnect urls on the given page and returns them.
+        """
+        disconnect_urls = []
+        for link_tag in page.soup.find_all('a'):
```
devl00p, line 387: You should use `page.links` here instead as it will directly provide absolute urls. Also you should maybe filter using `crawler.is_in_scope` to reduce false positives.

fwininger:
@JulienTD fix comment and rebase please.

PR #205, opened 2021-12-06
Added a new module to scan web technologies using the HashThePlanet database

JulienTD:
## Description

This pull request implements a new module to scan web technologies using the Hash The Planet database.  
This implementation doesn't any need external libraries. To be merged, it needs first to set up a link to download the database.  

## Related Issue(s)
 
N/A

### DIFF
--- a/wapitiCore/attack/mod_htp.py
+++ b/wapitiCore/attack/mod_htp.py
@@ -1,0 +1,183 @@
+import hashlib
+import json
+import os
+import re
+import sqlite3
+from typing import Dict, List, Tuple
+
+from httpx import RequestError
+from wapitiCore.attack.attack import Attack
+from wapitiCore.definitions.fingerprint_webserver import \
+    NAME as WEB_SERVER_VERSIONED
+from wapitiCore.language.vulnerability import _
+from wapitiCore.main.log import log_blue, logging
+from wapitiCore.net.page import Page
+from wapitiCore.net.web import Request
+
+MSG_TECHNO_VERSIONED = _("The range for {0} is from {1} to {2}")
+
+# types
+Technology = str
+Version = str
+
+class ModuleHtp(Attack):
+    """
+    Identify web technologies used by the web server using the HashThePlanet database.
+    """
+
+    name = "htp"
+
+    do_get = True
+    do_post = False
+    user_config_dir = None
+    finished = False
+    _db = None
+
+    HTP_DATABASE = "hashtheplanet.db"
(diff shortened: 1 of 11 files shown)


wapitiCore/attack/mod_htp.py
```diff
@@ -0,0 +1,183 @@
...
+    def __init__(self, crawler, persister, attack_options, stop_event):
+        Attack.__init__(self, crawler, persister, attack_options, stop_event)
+        self.tech_versions: Dict[Technology, List[Version]] = {}
+        self.user_config_dir = self.persister.CONFIG_DIR
+
+        if not os.path.isdir(self.user_config_dir):
+            os.makedirs(self.user_config_dir)
+
+    async def update(self):
+        """Update the Wappalizer database from the web and load the patterns."""
```
devl00p, line 48: s/Wappalizer/HashThePlanet/ :)

wapitiCore/attack/mod_htp.py
```diff
@@ -0,0 +1,183 @@
...
+                request=Request(root_url),
+                info=json.dumps(tech_info)
+            )
+            log_blue(MSG_TECHNO_VERSIONED, technology, truth_table[min_index], truth_table[max_index])
+        self._db.close()
+        self.finished = True
+
+    def _analyze_str(self, page_content: str) -> Tuple[str, str]:
+        cursor = self._db.cursor()
+        page_hash = hashlib.sha256(page_content.encode("utf-8")).hexdigest()
```
devl00p, line 143: You may have some encoding errors here. You should specify the `errors="ignore"` options but IMHO it would be better to directly hash the raw data using `response.bytes` instead of encoding from `response.content` this way you ensure to always have the correct hash even if the server gives an invalid charset

wapitiCore/attack/mod_htp.py
```diff
@@ -0,0 +1,183 @@
...
+        Retrieves the url's content and then analyze it to get the technology and the version
+        """
+        try:
+            response = await self.crawler.async_send(request, follow_redirects=True)
+        except RequestError:
+            self.network_errors += 1
+            return
+        if response.content is None or len(response.content) == 0:
+            return
+        content_hash = self._analyze_str(response.content)
```
devl00p, line 90: Maybe replace var name `content_hash` with something less missleading

devl00p:
Merging it. It is currently pretty slow for very few results though and I'm afraid it won't be scalable with a growing database.

I think it should be changed so:

- root url is fetched
- css / js / favicon urls are extracted from the root page
- hash is made for each of those resources
- hashes are looked up in the db

Using static paths may not work for some js files that are very popular but are stored at a variety of locations (wp-includes, assets, js, vendor, resources, etc) depending on the web application.

PR #206, opened 2021-12-10
[INSTALL.md] fix activation command for bash/sh

Kr1ss-XD:
I noticed a little inaccuracy in the instructions about using a virtual Python environment, in the section about activating the venv.

Please see the commit message.

_Thank you for maintaining and sharing the fabulous Wapiti by the way !_

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -36,7 +36,7 @@ python -m venv wapiti3
 Now let's activate it (make it our current working environment) :
  
 ```sh
-./wapiti3/bin/activate
+. wapiti3/bin/activate
 ```
 
 Or alternatively on Windows :


tarraschk:
Hello @Kr1ss-XD thank you for your contribution

Are you sure that you want to remove the slash from the command line? It would break most attempts to run the program.

Best regards

tarraschk:
OK we checked and you're right :)

Screenshot

tarraschk:
I'll my colleagues review as well the PR and merge. Thanks!

Kr1ss-XD:
Thx for your quick reply !

Since the previous command in these instructions was `python -m venv wapiti3`, i was thinking we can safely assume that the `wapiti3` venv is a subfolder of our current working directory.

Of course I'm fine with changing the command to

```sh
. ./wapiti3/bin/activate
```

but that would have the same effect.

For that matter, this command only _activates_ the virtual environment. During the activation, the `PATH` is altered, so that you can afterwards run Wapiti by just entering `wapiti <argument ...>`, without a path prefix.

devl00p:
Interesting @Kr1ss-XD 

It seemed to me it worked previously but maybe the venv module changed the default permission on those files. But the dot execution is common so the PR is good.

Thank you for your help

Kr1ss-XD:
Thank you very much.

PR #208, opened 2021-12-13
Now the wapp module will detect all the technologies on the root url and on the redirected one

JulienTD:
## Description
Before this pull request, the Wapp module was detecting technologies and their versions only on the redirection of the root URL.  
For instance, if the URL `https://test.com` redirects to `https://test.com/login/` and we decide to do an analysis on `https://test.com` it will in reality do an analysis on `https://test.com/login/`.
  
Sometimes, some headers are not transferred to the redirected URL, so we can lose precision on versions or in the worst case miss a technology.
  
In this pull request, the Wapp module is now doing an analysis on the root URL and also on the redirected URL. After that, it merges both results.  
  
The issue has been detected on this URL: https://216.176.178.100/owa/ the header `X-OWA-Version` is disappearing during the redirection.

## Related Issue(s)
N/A

### DIFF
--- a/wapitiCore/attack/mod_wapp.py
+++ b/wapitiCore/attack/mod_wapp.py
@@ -19,6 +19,7 @@ import json
 import os
 import asyncio
 import string
+from typing import Dict
 from httpx import RequestError
 
 from wapitiCore.main.log import logging, log_blue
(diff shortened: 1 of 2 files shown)


bretfourbe (approved):

wapitiCore/attack/mod_wapp.py
```diff
@@ -145,6 +139,42 @@ async def attack(self, request: Request):
...
+            detected_applications.append((
+                request,
+                wappalyzer.detect_with_versions_and_categories_and_groups()
+            ))
+
+        if not detected_applications:
+            return {}
+
+        # Merging the technologies & their versions
+        for request, detected_application in detected_applications:
```
jinxka, line 166: There should be an easier way to merge two dictionaries in python

jinxka (changes requested):

PR #215, opened 2021-12-15
Added a new exception to be caught in the function json

JulienTD:
## Description

Added a new exception to be caught in the function `json` of the class Page.

## Related Issue(s)
#219

### DIFF
--- a/wapitiCore/net/page.py
+++ b/wapitiCore/net/page.py
@@ -19,25 +19,24 @@
 
 # Standard libraries
 import re
-from functools import lru_cache
-from urllib.parse import urlparse, urlunparse
-from hashlib import md5
-from http.client import IncompleteRead
 import warnings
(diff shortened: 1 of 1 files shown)


wapitiCore/net/page.py
```diff
@@ -211,13 +211,13 @@ def json(self):
 
         try:
             return self._response.json()
-        except ValueError:
-            pass
+        except ValueError as e:
+            logging.exception(e)
```
devl00p, line 215: I'm not sure we should log the exception as we have no control on the server responses.

If a server goes mad (5XX errors like internal or gateway errors) during scan I'm afraid it will generate way too much exceptions.

devl00p (changes requested):

devl00p (approved):

PR #226, opened 2022-01-02
Rework packaging so Wapiti can access its resources from the egg file

devl00p:


wapitiCore/wappalyzer/wappalyzer.py
```diff
@@ -33,9 +30,10 @@ def __init__(self, categories_file_path=None, groups_file_path=None, technologie
         data_filename : str
             File providing application and categorie references (Json format).
         """
-        default_categories_file_path = os.path.join(BASE_DIR, "wappalyzer", "data/categories.json")
```
devl00p, line 36: @JulienTD Those files were searched in the package directory while they aren't versionned. Was it intended ?

As they would fail to load with an egg file I had to change the default location.

If you are preloading the files they should be put in the user's `.wapiti` directory instead.

JulienTD: Hello! Currently these fles are not versioned but they can be updated (they depend on this repository: https://github.com/wapiti-scanner/wappalyzer), so they need to stay in the .wapiti folder.  
Maybe we can version them, and tell the user if an update is needed.

JulienTD (approved):

PR #230, opened 2022-01-10
Add Django’s CSRF HTML form name to the token list

thomasjbradley:
Allow Wapiti to confirm Django has submitted the correct anti-CSRF form POST field.

This setting is not configurable within Django, so there’s no way for a developer to change the field’s name to match one of the tokens in the list.


devl00p:
Thank you for this contribution

PR #235, opened 2022-01-19
Added a new flag to show the detail of the requests and the responses

### LINKED ISSUE #37: Report Full HTTP Request and Response
To better understand an issue it would be great if the final report contains the **full** HTTP request **and** response.
This would be great a feature to better understand and help fixing found issues within an application.

JulienTD:
## Description
Currently, when we use Wapiti, it will generate a report under different formats: HTML, JSON, XML, and  TXT.  
This report will contain information about the found vulnerabilities/additionals/anomalies:
- the HTTP method
- the path
- information about the payload
- the depth of the request on the website
- the used parameter
- the referer
- the module
- some information about the HTTP request (URL, referer, params)
- the curl command
- the WSTG code
  
But we don't have any detailed information about the request and the response.  
In this pull request, I have added a new flag `--detailed-report` or `-dr` to have the missing data on the generated reports.  
It will create a new section on each payload containing the detailed version of the request and the response.  
The request will have:  
- the URL
- the method
- the headers
- the query
- the body
- the encoding
  
The response will have:
- the status code
- the body
- the headers
  
Also to have a deeper understanding of the scan that we are doing, every sent request will be saved under the `infos` section of the report file.  
The number of crawled pages will now be written under the field `crawled_pages_nbr` and the crawled pages under the field `crawled_pages`.
  

To do this feature, I had to modify the `add_payload` function signature to have a new argument: `response`, to save it after in the payload. I had also to modify the `async_explore` function to return the request and the response to also save it in the payload.  
There is also a new table named `Response` that is linked to the `Payload` and `Path` tables.

## Related Issue(s)
fixes #37


JulienTD:
ping @bretfourbe

devl00p (commented):
Congratulation for that stuff :)

Lot of cool things. There are still a few things that I'd like in another way and I will make you a PR at some point. For exemple I found a way to get the full headers actually sent by httpx and made it mandatory (they will appear in the `http_repr` along with cookies).

About how to store and display the data I chose the `LargeBinary` data type and to display I simply decode as UTF-8 with `errors="replace"`. As most of the vulnerable responses should be HTML and error messages in english that seems enough. If there are some binary responses then the unsupported characters will be replaced.

Now the problematic part :stuck_out_tongue_closed_eyes: 

While I see a benefit at seing the responses for vulnerable webpages (people can me sure there is no false positives, copy data for reports, understand how Wapiti works, etc (that's why the responses body should not appear in base64 to the user)) I don't think it is useful to store the responses for all crawled webpages.

Also there are currently no size limit on HTTP responses in Wapiti for the moment so you may end up with a response of several Gb or several big pages that will have an impact on Wapiti performance.

If you have important reasons to keep that feature then tell me. We would then use it only with `--detailed-report` and other things would be mandatory (full requests/responses for vulnerabilities). Otherwise I think it should be removed.

devl00p:
@JulienTD I didn't find a way to make a PR to your fork (don't know if it is possible) but it looks like I was able to push something on your branch :)

I didn't spend time on the tests yet but you can already take a look at the changes

wapitiCore/net/sql_persister.py
```diff
@@ -17,18 +17,25 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 import json
+import logging
```
bretfourbe, line 20: This should not be necessary since there is logging from `wapitiCore.main.log`

PR #237, opened 2022-01-20
Upgrade httpx to latest (0.21.3)

devl00p:


Pipfile
```diff
@@ -10,7 +10,7 @@ pytest-asyncio = "==0.14.0"
 pytest-cov = ">=2.11.1"
 
 [packages]
-"httpx[brotli]" = "==0.21.1"
```
fwininger, line 13: It would be nice to do the same in setup.py to avoid breakages.

PR #243, opened 2022-02-07
Changed the payload type for the technology detection to addition instead of vulnerability

JulienTD:
## Description
Currently, when technology is found a vulnerability payload will be saved, but the problem is that some modules will save a payload as an addition, so for the same content, it will be split into two different categories.  
In this pull request, I have changed the type of payload for all the technology detection to an addition.
Also, when a version is found it is registered as a vulnerability, and when a technology is found it is registered as an addition.

## Related Issue(s)
N/A

### DIFF
--- a/tests/attack/test_mod_wp_enum.py
+++ b/tests/attack/test_mod_wp_enum.py
@@ -109,16 +109,23 @@ async def test_plugin():
     await module.attack(request)
 
     assert persister.add_payload.call_count
+
     assert persister.add_payload.call_args_list[0][1]["module"] == "wp_enum"
     assert persister.add_payload.call_args_list[0][1]["category"] == _("Fingerprint web technology")
     assert persister.add_payload.call_args_list[0][1]["info"] == (
-        '{"name": "bbpress", "versions": ["2.6.6"], "categories": ["WordPress plugins"]}'
+        '{"name": "WordPress", "versions": [], "categories": ["CMS", "Blogs"], "groups": ["Content"]}'
     )
+
+    assert persister.add_payload.call_args_list[1][1]["module"] == "wp_enum"
+    assert persister.add_payload.call_args_list[1][1]["category"] == _("Fingerprint web technology")
     assert persister.add_payload.call_args_list[1][1]["info"] == (
-        '{"name": "wp-reset", "versions": [""], "categories": ["WordPress plugins"]}'
+        '{"name": "bbpress", "versions": ["2.6.6"], "categories": ["WordPress plugins"], "groups": ["Add-ons"]}'
     )
     assert persister.add_payload.call_args_list[2][1]["info"] == (
-        '{"name": "unyson", "versions": [""], "categories": ["WordPress plugins"]}'
+        '{"name": "wp-reset", "versions": [""], "categories": ["WordPress plugins"], "groups": ["Add-ons"]}'
+    )
+    assert persister.add_payload.call_args_list[3][1]["info"] == (
+        '{"name": "unyson", "versions": [""], "categories": ["WordPress plugins"], "groups": ["Add-ons"]}'
     )
     await crawler.close()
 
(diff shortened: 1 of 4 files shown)


wapitiCore/attack/mod_wapp.py
```diff
@@ -126,14 +126,14 @@ async def attack(self, request: Request):
 
             if versions:
                 if "Web servers" in categories:
-                    await self.add_vuln_info(
+                    await self.add_vuln(
```
bretfourbe, line 129: use `add_vuln_info`

wapitiCore/attack/mod_wapp.py
```diff
@@ -126,14 +126,14 @@ async def attack(self, request: Request):
...
                         category=WEB_SERVER_VERSIONED,
                         request=request_to_root,
                         info=json.dumps(detected_applications[application_name]),
                         wstg=WEB_SERVER_WSTG_CODE
                     )
                 else:
-                    await self.add_vuln_info(
+                    await self.add_vuln(
```
bretfourbe, line 136: idem

wapitiCore/attack/mod_wp_enum.py
```diff
@@ -70,22 +69,27 @@ async def detect_version(self, url: str):
...
+
+        log_blue(
+            MSG_WP_VERSION,
+            "N/A" if not detected_version else detected_version
+        )
+        info_content = {"name": "WordPress", "versions": [], "categories": ["CMS", "Blogs"]}
+
+        if detected_version:
+            info_content["versions"].append(detected_version)
+            await self.add_vuln(
```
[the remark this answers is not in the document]
JulienTD: Sorry for the MR not being very understandable, I will modify the description to make it better.  
Concerning the code, when a version is found it is registered as a vulnerability and if the technology is found then we register it as an addition, that's why in this case there are two saved payloads: a vulnerability for the version and an addition for the technology.

wapitiCore/attack/mod_wp_enum.py
```diff
@@ -70,22 +69,27 @@ async def detect_version(self, url: str):
...
+
+        log_blue(
+            MSG_WP_VERSION,
+            detected_version or "N/A"
+        )
+        info_content = {"name": "WordPress", "versions": [], "categories": ["CMS", "Blogs"]}
+
+        if detected_version:
+            info_content["versions"].append(detected_version)
+            await self.add_vuln(
```
bretfourbe, line 81: use àdd_vuln_info` here

tests/attack/test_mod_wp_enum.py
```diff
@@ -109,15 +109,15 @@ async def test_plugin():
     await module.attack(request)
 
     assert persister.add_payload.call_count
-    assert persister.add_payload.call_args_list[0][1]["module"] == "wp_enum"
```
bretfourbe, line 112: please add a test case for additional

wapitiCore/attack/mod_wp_enum.py
```diff
@@ -70,22 +69,27 @@ async def detect_version(self, url: str):
...
+                category=TECHNO_DETECTED,
```
bretfourbe, line 82: category should be `WEB_APP_VERSIONED`

wapitiCore/attack/mod_drupal_enum.py
```diff
@@ -137,17 +138,25 @@ async def attack(self, request: Request):
...
-                "categories": ["CMS Drupal"]
+                "categories": ["CMS Drupal"],
+                "groups": ["Content"]
             }
             log_blue(
                 MSG_TECHNO_VERSIONED,
                 "Drupal",
                 self.versions
             )
+            if len(self.versions) > 0:
```
bretfourbe, line 153: replace with `if self.versions:`

JulienTD:
ping @bretfourbe

devl00p (approved):
looks good to me

PR #255, opened 2022-02-16
Added tests to the authentication functions

JulienTD:
## Description
In this pull request, I have added tests for the authentication functions and I have also improved the readability of the function `_async_try_login_post`.

## Related Issue(s)
N/A


tests/net/test_crawler.py
```diff
@@ -72,3 +74,291 @@ def test_extract_disconnect_urls_multiple_urls():
...
+
+    assert is_logged_in is False
+    assert len(form) == 0
+    assert len(disconnect_urls) == 0
+
+    is_logged_in, form, disconnect_urls = await crawler._async_try_login_basic_digest_ntlm(auth_url_3)
+
+    assert is_logged_in is False
+    assert len(form) == 0
+    assert len(disconnect_urls) == 0
```
polyedre, line 325: You can use a for loop here

polyedre: In fact, you can use a for loop for the whole test

polyedre (approved):

tests/net/test_crawler.py
JulienTD: Done :rocket:

polyedre (approved):

PR #258, opened 2022-02-16
WIP: Abstract the concept of a vulnerability inside its own class

polyedre:
This is a work-in-progress to gather feedback.

The current idea is to replace progressively the data inside the `definitions` directory with classes inherited from a generic `Vulnerability` class. This approach allow vulnerabilities to specify additional fields with explicit names instead of using the `info` field.

Each class inheriting from the Vulnerability class represent a type of vulnerability and can have a name, short_name, description, solution and wstg codes. Each instance of a class inherited from the Vulnerability class represent a vulnerability detected in the target.


polyedre:
In the function `add_payload` in file `attack.py`, what is the request in the `request` argument?

```python
    async def add_payload(self, payload_type: str, category: str, request_id: int = -1,
                          level=0, request=None, parameter="", info="", wstg=None):
        await self.persister.add_payload(
            request_id=request_id,
            payload_type=payload_type,
            module=self.name,
            category=category,
            level=level,
            request=request,
            parameter=parameter,
            info=info,
            wstg=wstg
        )
```

It seems that `request_id` always refers to the `Request` sent as a parameter of the `attack` method of the module, and when `request` is used, it represents the request that validates the existence of the vulnerability. Is it true @bretfourbe  @devl00p?

devl00p:
@polyedre The `request` parameter must indeed be the Wapiti `Request` object corresponding to the HTTP request using the attack payload.

The `request_id` parameter should match the Wapiti `Request.request_id` for the original request that the crawler found (without any payload)

Issue #256, opened 2022-02-16
Scan a specific POST request


devl00p:
sqlmap provides a feature to attack a specific request by loading it from a dump:

```
-r REQUESTFILE      Load HTTP request from a file
```

Currently, playing with Wapiti parameters it is possible to do the same thing on a GET request:

`-u <url> --scope url -H "Header: etc"`

But we have no such things for POST requests.

Proposal: add a `--data` option to specify POST parameters like this

```--data 'id=5&file=profile.php'```

This feature would be allowed to work whatever the scope used but we should be careful that scope checking doesn't break due to referring a POST request (I don't think it would be the case)

PR #260, opened 2022-02-20
Allows to perform an attack on a single POST request.

devl00p:
Parameters given using a `--data` options to Wapiti

Fix https://github.com/wapiti-scanner/wapiti/issues/256

### DIFF
--- a/tests/attack/test_mod_backup.py
+++ b/tests/attack/test_mod_backup.py
@@ -24,7 +24,7 @@ async def test_whole_stuff():
     request.path_id = 1
     request.set_headers({"content-type": "text/html"})
 
-    crawler = AsyncCrawler("http://perdu.com/", timeout=1)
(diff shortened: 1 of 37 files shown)


devl00p:
capt

JulienTD (commented):
LGTM. I was wondering if we could also apply this feature to the starting URLs if they are provided by the user. For example, we could maybe use this feature to authenticate the user to a website using a POST authentication system.

devl00p:
> LGTM. I was wondering if we could also apply this feature to the starting URLs if they are provided by the user. For example, we could maybe use this feature to authenticate the user to a website using a POST authentication system.

I'm afraid it would be difficult for users to correctly understand how to use the option if it seems to apply on both the base and starting URLs.

The `wapiti-getcookie` offers the `--data` option solely for authentication so for now users should either use the `-s` with 'post' auth or the `wapiti-getcookie` utility.

PR #263, opened 2022-02-21
Replaced hard-coded status code to a wrapper function

### LINKED ISSUE #253: Update success checking
To check whether a request was successful or not, we currently do operation on the status_code like `response.code != 400` or `response.code == 200`.

We should remove all hard coded return code and use `response.is_success` instead.

see https://www.python-httpx.org/compatibility/#checking-for-success-and-failure-responses

JulienTD:
## Description
Currently, Wapiti uses hardcoded status codes to verify if a request is successful or not.  
In this pull request, I have added 5 methods to determine if the status of a request is:  
- successful (2xx status codes)
- redirect (3xx status codes)
- error (4xx and 5xx status codes)
- client_error (4xx status codes)
- server_error (5xx status codes)

I have replaced these hardcoded status codes with these methods.

## Related Issue(s)
fixes #253

### DIFF
--- a/wapitiCore/attack/mod_backup.py
+++ b/wapitiCore/attack/mod_backup.py
@@ -61,7 +61,7 @@ class ModuleBackup(Attack):
                 # Do not put anything in false_positive_directories, another luck for next time
                 return False
 
-            self.false_positive_directories[request.dir_name] = (response and response.status == 200)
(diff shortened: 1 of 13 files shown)


devl00p:
Looks good to me

polyedre (approved):

PR #289, opened 2022-06-06
Rename wapitiCore.net.page.Page to wapitiCore.net.response.Response

devl00p:


devl00p:
The idea is to later move HTML stuff from Response to a new class (`Html`)

PR #304, opened 2022-07-22
Added iframe and anchor src/href direct control

un-fmunozs:
Additional XSS detection, two cases that I've seen ITW.

```
<a href="CONTROL">anything</a>
<iframe src="CONTROL"></iframe>
```

xss_utils change was made in a way to avoid doing a lot of touching on that section. comments welcome.


meli-femunoz:
Hello team, is there anything missing from this PR?

devl00p:
changed the check a bit so it wont allow  <a src=something /> and also moved the "a src" payload at the end of the ini file because it requires interaction (if we can get any payload withoit interaction it should be used in proority)

PR #311, opened 2022-07-27
Fix crash when missing environment variables

undergroundwires:
Fixes crash when wapiti is called without `HOME` and `USERPROFILE` environment variables. It is caused by `os.path.joins` when a home directory is None.

Reproduce using: `env -i wapiti --version` > gives error

It can be resolved by providing HOME environment variable such as:

`env -i HOME=/root wapiti --version` > prints version.

Environment variables can be missing due to different isolation approaches such as a slim Docker image or an isolated shell.

This PR/commit fixes the issue by defaulting to `/home` directory that's Linux/Unix standard when an environment variable is missing. This way the crash is avoided.

### DIFF
--- a/tests/attack/test_mod_htp.py
+++ b/tests/attack/test_mod_htp.py
@@ -19,7 +19,7 @@ from wapitiCore.net import Request
 @respx.mock
 async def test_must_attack():
     persister = AsyncMock()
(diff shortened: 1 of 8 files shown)


devl00p:
Merged, thank you

PR #312, opened 2022-07-27
Feature/openredirect doubleslash

un-fmunozs:
Detect open redirects when full urls with schema are not allowed, this uses //openbugbounty.org as a payload.

The detection mechanism had to be updated because when the scanned site is HTTP, the redirect will also be to HTTP and  if we only check HTTPS://openbugbounty.org it will not detect the openredirect.

Test cases added

### DIFF
--- a/tests/attack/test_mod_redirect.py
+++ b/tests/attack/test_mod_redirect.py
@@ -46,6 +46,24 @@ async def test_redirect_detection():
         ]
 
 
+
+@pytest.mark.asyncio
+async def test_redirect_detection_no_url():
+    persister = AsyncMock()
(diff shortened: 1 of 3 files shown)


un-fmunozs:
Hello team, I sent this PR and https://github.com/wapiti-scanner/wapiti/pull/304 a while ago, and would love to work on more features for wapiti. Is there something I am missing to get this reviewed and accepted or rejected? Thanks.

bretfourbe:
Hi @un-fmunozs, thank you for the contribution :) it looks good, what about adding `@openbugbounty.org` and other bypass paylaods ? :) 
@devl00p

devl00p:
thank @un-fmunozs for your help :)

I'll take a look at the other PR

Regards

Issue #328, opened 2022-10-06
Rework authentication


devl00p:
There are several issues with auth in current Wapiti:

- we are using a tuple for creds which is not very pythonic
- we are using the same struct and the same option for either basic/ntlm/login-through-html-form which is not good as a form may be submitted while basic or ntlm auth are in use
- the `-s` option for form logins is not intuitive
- we can't pass some stupid-yet-problematic forms that submit creds in urlencoded form but some fields can't be found directly (we need to integrate more getcookie logic inside Wapiti)
- if `--skip-crawl` is in use we don't check credentials with  `async_try_login` while we should do it in every case, especially for form-logins because in that case we do not simply check creds but we are actually login in and obviously we want to login again because web creds may have expired

Proposal:

- use distinct fields for username and password
- separate structs and options for basic/ntlm auth on one side and form auth on the other side (add new specific options)
- use a new option for the login URL (replace the `-s` for that usage) and another option for the enctype (if not set consider urlencoded). we may use a different behavior when the enctype option is used (Without it, consider the URL to be the one with the form, parse it and fill it. With enctype set, consider the URL to be the target of the form, send data directly)
- use `async_try_login` even if `--skip-crawl` is set
- don't deprecate getcookie as at it is still handy for (for example) generating an empty cookie file before editing it manually and users may be used to it

devl00p:
Fixed

PR #331, opened 2022-10-17
Rework of authentication: separate HTTP auth from form based auth

devl00p:
+ adds some new options (not yet implemented)


devl00p:
Fix #328

PR #333, opened 2022-10-23
fix checking of geckodriver presence

devl00p:


devl00p:
geckodriver presence was checked and a warnign message appeared in some cases but we still were using args.headless at other places

PR #337, opened 2022-10-23
upgrade dnspython

devl00p:


bretfourbe:
Hi @devl00p, I think tests in log4shell passed due to a lack of verification from dnspython : 
image
This line is not allowed anymore https://github.com/wapiti-scanner/wapiti/blob/8a48d183a578d579513b77a6a4b4f3c72a6b4cfb/tests/attack/test_mod_log4shell.py#L123
We should use an IP address instead, and we should specify it in the documentation since it is directly linked to `--dns-endpoint` option value.

PR #343, opened 2022-10-30
add payload for loknop lfi to rce technique

devl00p:
https://gist.github.com/loknop/b27422d355ea1fd0d90d6dbc1e278d4d


devl00p:
great technique @loknop

PR #350, opened 2022-11-23
Dependencies reduction

Darkiros:
Reduce dependencies to the minimum required.


pyproject.toml
```diff
@@ -0,0 +1,91 @@
...
+        "pytest-cov>=2.11.1",
+        "pytest-asyncio==0.20.1",
+        "respx==0.20.0", 
+        "sslyze==5.0.6"
+]
+
+ssl = [
+        "humanize==4.4.0",
+        "sslyze==5.0.6"
+]
```
Darkiros, line 76: @polyedre what do you think about it ? Is it better to install ssl module like that ?

polyedre (commented):
I am not convinced by these changes:
1. You duplicated the packages and versions in one additional file. Two files are already too much, but 3... (requirements.txt, pyproject.toml and Pipfile)
2. You added a lot of explicit packages. This may cause dependency headaches in the future. And that's already the case ;)

If this change can only save about 10MB, I think it is not worth increasing the maintenance cost.

PR #362, opened 2023-01-09
fix auth creds option handling

devl00p:
should fix handling of -a -and --form-cred options following refactoring

### DIFF
--- a/tests/cli/test_options.py
+++ b/tests/cli/test_options.py
@@ -1,9 +1,10 @@
 import sys
 from asyncio import Event
 from unittest import mock
-from httpcore import URL
 
+from httpcore import URL
 import pytest
+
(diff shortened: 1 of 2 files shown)


tarraschk:
@devl00p merci pour la PR !

@bretfourbe c'est suite au signalement du bug que tu m'as indiqué sur les creds, pourrais-tu regarder si ça corrige ?

devl00p:
@bretfourbe Indeed :| I tried to add the `pytest.mark.xfail` on the mod_ssl test but it seems that flag isn't taken into consideration in the Docker env, can't figure out why at the moment

devl00p:
fixed pydantic version to satisfy sslyze and bumped some dependencies on master

PR #377, opened 2023-02-20
Remove port from Request#_ressource_path

jinxka:
To test the fix I used the following commands:

wapiti -u http://testphp.vulnweb.com:80 -m wp_enum --flush-session -v2

wapiti -u http://testphp.vulnweb.com -m wp_enum --flush-session -v2

and I made sure that both command scan the same number of pages which should be 76:

[*] Wapiti found 76 URLs and forms during the scan


wapitiCore/net/web.py
```diff
@@ -382,6 +382,10 @@ def __init__(
         self._start_time = None
         self._size = 0
         self._path_id = None
+        # Remove port from ressource_path
+        parsed_url = urlparse(self._resource_path)
+        self._resource_path = urlunparse((parsed_url.scheme, parsed_url.netloc.split(':')[0], parsed_url.path,
```
fwininger, line 387: supprimer la ligne précédente et:

```python
self._resource_path = urlunparse(parsed.scheme, parsed.hostname, parsed.path)
```

jinxka:
@devl00p Hello, I can't figure out how to fix the specs, I tried a few things but I always get an error:

`httpx.ConnectError: All connection attempts failed`

Issue #376, opened 2023-02-20
Error when specifying a port in an URL


jinxka:
I noticed that the scan doesn't work when a port is specified in an URL.

For example a wapiti scan on 'http://testphp.vulnweb.com/' finds 76 urls while the same scan on 'http://testphp.vulnweb.com:80/' finds only 1 url.

I checked Request#port and it has the right value (80 with both urls).

fwininger:
@devl00p could you take a look ?

devl00p:
I'm on it

devl00p:
should be ok, closing the issue

PR #379, opened 2023-02-23
remove unnecessary ports from URLs 

devl00p:
for example in https://www.google.com:443/ or http://perdu.com:80/


devl00p:
fix #376

Issue #385, opened 2023-03-17
Unable to import module ssl


Brian1619:
Hello, I have tried to use the Dockerfile in the repositiry to get a working env up and running. After building the image and trying to scan a site, Wapiti always says: 

[!] Unable to import module ssl
[!] Unable to find a module named ssl

When I connect to the container, and run `python3 -c "import ssl; print(ssl.OPENSSL_VERSION)"` 
I get `OpenSSL 1.1.1n  15 Mar 2022`

I started on a mac and had the same Problem. I build another Debian Docker image and tried install with `pip install wapiti3`, same results. 

Any tips on how to get Wapiti3 up and running?

bretfourbe:
Hi @Brian1619,

Can you have a look at https://github.com/wapiti-scanner/wapiti/issues/365 ?

Brian1619:
Hello @bretfourbe,
yes, I already saw that and tried what I found in #365. I have tried multiple ways to get Wapiti up and running. 
I tried in a fresh install and with a virtual Env. its always the same problem. What surprises me the most is that the Dockerfile setup has the same problem. 

We are trying to scan a https only application. I haven't found anything that says Wapiti can't handle that.

here is the my virtual env: 

(wapiti3) root@c7996e05cf69:/usr/src/app# pip freeze
aiocache==0.12.0
aiohttp==3.8.4
aiosignal==1.3.1
aiosqlite==0.17.0
anyio==3.6.2
arsenic==21.8
asgiref==3.5.2
async-timeout==4.0.2
attrs==22.2.0
beautifulsoup4==4.11.2
Brotli==1.0.9
browser-cookie3==0.16.2
certifi==2022.12.7
cffi==1.15.1
charset-normalizer==3.1.0
click==8.1.3
cryptography==38.0.4
dnspython==2.1.0
Flask==2.2.3
frozenlist==1.3.3
greenlet==2.0.2
h11==0.14.0
h2==4.1.0
hpack==4.0.0
httpcore==0.16.3
httpx==0.23.3
hyperframe==6.0.1
idna==3.4
importlib-metadata==6.0.0
itsdangerous==2.1.2
jaraco.classes==3.2.3
jeepney==0.8.0
Jinja2==3.1.2
kaitaistruct==0.10
keyring==23.13.1
ldap3==2.9.1
loguru==0.6.0
lz4==4.3.2
Mako==1.2.4
MarkupSafe==2.1.1
mitmproxy==9.0.0
[...]

PR #387, opened 2023-03-27
put "ssl" extra dependencies in pyproject 

devl00p:
+ add more info on module import error

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,6 +62,10 @@ wapiti = "wapitiCore.main.wapiti:wapiti_asyncio_wrapper"
 wapiti-getcookie = "wapitiCore.main.getcookie:getcookie_asyncio_wrapper"
 
 [project.optional-dependencies]
+ssl = [
+        "humanize==4.4.0",
+        "sslyze==5.1.1"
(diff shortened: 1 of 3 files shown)


devl00p:
related to #385

pyproject.toml
```diff
@@ -62,6 +62,10 @@ wapiti = "wapitiCore.main.wapiti:wapiti_asyncio_wrapper"
 wapiti-getcookie = "wapitiCore.main.getcookie:getcookie_asyncio_wrapper"
 
 [project.optional-dependencies]
+ssl = [
+        "humanize==4.4.0",
+        "sslyze==5.1.1"
```
bretfourbe, line 67: I am not sure but maybe you also need `pydantic` ?

bretfourbe: @devl00p so this line `"pydantic==1.10.2" # This is to avoid a conflict with sslyze` is not needed anymore in `test` dependencies just below ?

devl00p: It seems it works with recent versions of pydantic (1.10.7) so it should be ok

PR #388, opened 2023-04-04
add flexible timeout for headless crawl

thr3athunt3r:
Hard-coded timeouts can sometimes raise exceptions.

### DIFF
--- a/wapitiCore/controller/wapiti.py
+++ b/wapitiCore/controller/wapiti.py
@@ -156,7 +156,6 @@ class Wapiti:
 
         self.crawler_configuration = CrawlerConfiguration(self.base_request)
         # self.crawler = None
-
         self.target_scope = Scope(self.base_request, scope)
 
         self.report_gen = None
@@ -187,6 +186,7 @@ class Wapiti:
         self.detailed_report = False
         self._headless_mode = "no"
         self._wait_time = 2.
+        self._timeout = 6.
 
         if session_dir:
             SqlPersister.CRAWLER_DATA_DIR = session_dir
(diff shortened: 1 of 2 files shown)


bretfourbe:
Hi @thr3athunt3r, can you create an issue to explain the problem and give examples and what behavior do you expect from wapiti ?
Here it seems that timeout is still hard-coded, you just changed the value of timeout.

wapitiCore/controller/wapiti.py
```diff
@@ -187,6 +186,7 @@ def __init__(self, scope_request: Request, scope="folder", session_dir=None, con
         self.detailed_report = False
         self._headless_mode = "no"
         self._wait_time = 2.
+        self._timeout = 6.
```
devl00p, line 189: `_timeout` should not be declared in the `Wapiti` class as a timeout is already stored in the `CrawlerConfiguration` object.

wapitiCore/net/intercepting_explorer.py
```diff
@@ -379,7 +379,8 @@ def __init__(
             drop_cookies: bool = False,
             headless: str = "no",
             cookies: Optional[CookieJar] = None,
-            wait_time: float = 2.
+            wait_time: float = 2.,
+            timeout: float = 6.,
```
devl00p, line 383: No need for this argument as you are using `crawler_configuration.timeout`

Note that you should be able to directly get the timeout using `self._crawler.timeout`

wapitiCore/net/intercepting_explorer.py
```diff
@@ -319,14 +319,14 @@ async def launch_headless_explorer(
 
                 if request.method == "GET":
                     try:
-                        await headless_client.get(request.url_with_fragment, timeout=5)
+                        await headless_client.get(request.url_with_fragment, timeout=timeout)
```
devl00p, line 322: I think it can be simplified again using `crawler.timeout`.

bretfourbe:
Hi @thr3athunt3r,
I think this MR fix your issue: https://github.com/wapiti-scanner/wapiti/pull/422 ?

PR #393, opened 2023-04-19
Remove allow-from directive

danielklecha:
Allow-from is an obsolete directive that no longer works in modern browsers.

### DIFF
--- a/wapitiCore/attack/mod_http_headers.py
+++ b/wapitiCore/attack/mod_http_headers.py
@@ -33,7 +33,7 @@ INFO_XFRAME_OPTIONS = "X-Frame-Options is not set"
 class ModuleHttpHeaders(Attack):
     """Evaluate the security of HTTP headers."""
     name = "http_headers"
-    check_list_xframe = ['deny', 'sameorigin', 'allow-from']
+    check_list_xframe = ['deny', 'sameorigin']
     check_list_xcontent = ['nosniff']
     check_list_hsts = ['max-age=']
 
--- a/wapitiCore/attack/mod_http_headers.py
+++ b/wapitiCore/attack/mod_http_headers.py
@@ -25,6 +25,13 @@ from wapitiCore.main.log import log_blue, log_green, log_orange, log_red
 from wapitiCore.net.response import Response
 from wapitiCore.net import Request
 
+HSTS_NOT_SET = "Strict-Transport-Security is not set"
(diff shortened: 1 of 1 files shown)


bretfourbe:
Hi @danielklecha, thanks for contributing.
Indeed, according to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options, `allow-from` is now deprecated.
What about raising the use of a deprecated value instead of just removing it ?
@devl00p

danielklecha:
@bretfourbe Maybe is_set method should be split into is_set and contains. This could give 3 states:
- is not set => error
- doesn't contain any element from checklist => warning "X-Frame-Options contains invalid value"
- else => success

danielklecha:
@bretfourbe I updated the code and tested it on my machine. Is it ok?

bretfourbe:
I would keep variables for these strings
```python
HSTS_NOT_SET = "Strict-Transport-Security is not set"
XCONTENT_TYPE_NOT_SET = "X-Content-Type-Options is not set"
XFRAME_OPTIONS_NOT_SET = "X-Frame-Options is not set"
``` 
and add the following ones
```python
INVALID_HSTS = "Strict-Transport-Security has an invalid value"
INVALID_XCONTENT_TYPE = "X-Content-Type-Options has an invalid value"
INVALID_XFRAME_OPTIONS = "X-Frame-Options has an invalid value"
``` 
And in check_list change `info` field like this ?
```python
info = {"error": HSTS_NOT_SET, "warning": INVALID_HSTS}
``` 
So you don't need to change `check_header` parameters ?

danielklecha:
@bretfourbe I updated PR based on your suggestions.

tarraschk (approved):

tarraschk:
The code looks okay for me, thanks for your contribution

@bretfourbe peux-tu regarder si le format de sortie reste identique à ce qu'on produisait auparavant ? C'est important pour qu'on récupère correctement les données dans Cyberwatch

bretfourbe:
Hi @danielklecha, looks good to me !
Can you just fix the coding style issue `wapitiCore/attack/mod_http_headers.py:68:0: C0303: Trailing whitespace (trailing-whitespace)` please ?
Thanks

danielklecha:
@bretfourbe done
Thanks!

Issue #374, opened 2023-02-15
Add support for uncommon http methods


juansec:
This is an idea, not a current issue:

It would be nice if it is possible to change the default GET/POST HTTP methods to uncommon methods such as PUT, DELETE, etc.

Benefits:

- It would be useful for API testing

devl00p:
Currently there is the `methods` module in Wapiti that performs an OPTIONS HTTP request to see if uncommon methods like PUT or DELETE are allowed. It may satisfy your needs but it won't work if the server doesn't support OPTIONS.

I think the module could be improved to check OPTIONS first then attempt PUT based on what the response was.

PR #395, opened 2023-04-26
improving methods module and tests about it

RMI78:
This PR is an answer for the issue #374 
It allows testing uncommon methods when they are detected with the OPTIONS method. Only methods with a different body content or a different server return code than the GET method will be listed. Methods returning a 405 error code (method not allowed) will also be ignored.

```Bash
wapiti -u https://misbehaving.site/anything -m methods --flush-session  

 ██╗    ██╗ █████╗ ██████╗ ██╗████████╗██╗██████╗
 ██║    ██║██╔══██╗██╔══██╗██║╚══██╔══╝██║╚════██╗
 ██║ █╗ ██║███████║██████╔╝██║   ██║   ██║ █████╔╝
 ██║███╗██║██╔══██║██╔═══╝ ██║   ██║   ██║ ╚═══██╗
 ╚███╔███╔╝██║  ██║██║     ██║   ██║   ██║██████╔╝
  ╚══╝╚══╝ ╚═╝  ╚═╝╚═╝     ╚═╝   ╚═╝   ╚═╝╚═════╝  
Wapiti 3.1.7 (wapiti-scanner.github.io)
[*] Saving scan state, please wait...

[*] Launching module methods
---
Interesting methods allowed on https://misbehaving.site/anything: DELETE, PATCH, PUT
Method DELETE returned a body content different from GET method on https://misbehaving.site/anything
Method PUT returned a body content different from GET method on https://misbehaving.site/anything
Method PATCH returned a body content different from GET method on https://misbehaving.site/anything
---
```

Unit tests were also adapted and improved to support this new behavior.

### DIFF
--- a/wapitiCore/attack/mod_methods.py
+++ b/wapitiCore/attack/mod_methods.py
@@ -34,10 +34,66 @@ class ModuleMethods(Attack):
 
     name = "methods"
     PRIORITY = 6
-    KNOWN_METHODS = {"GET", "POST", "OPTIONS", "HEAD", "TRACE"}
+    KNOWN_METHODS = {"GET", "POST", "OPTIONS", "HEAD", "TRACE", }
+    UNCOMMON_METHODS = {"CONNECT", "DELETE", "PUT", "PATCH"}
     do_get = True
     do_post = True
     excluded_path = set()
+
+    async def discover_methods(self, request: Request, response: Response, methods: set):
+        """
+        A function to try the various methods returned by an OPTIONS method and comparing them
+        with the method used by the crawler to see the differences.
+        If OPTIONS returned nothing (empty methods) or a 404 status code,
+        it will blindly try all the known methods. A fully empty OPTIONS
+        methods seems suspicious so it's better to ultimately check everything.
+
+        Parameter List
+        -------------
+        request : Request
+            Request made previously, useful for its informations in order to make new requests
+        response : Response
+            Response from the request argument, used to compare to responses from requests made with methods
+        methods: set
+            set of all methods returned by an OPTIONS method request
+
+        Return object
+        -------------
+        methods_dict : dict
+            dictionnary referencing relevant methods (not 405 status code, differences with the response in argument)
+            and their informations
+
+        """
+        methods_dict = {}
+        # Testing and comparing the uncommon methods,
+        # not_allowed_methods record all the methods
+        # returning a 405 error which will be removed
+        all_but_options_method = set.union(self.KNOWN_METHODS, self.UNCOMMON_METHODS)-{'OPTIONS'}
+        for method in (methods-self.KNOWN_METHODS if methods else all_but_options_method):
+            uncommon_request = Request(
+                request.path,
+                method,
+                referer=request.referer,
+                link_depth=request.link_depth
+            )
+            try:
+                log_verbose(f"[+] {uncommon_request}")
+                uncommon_response = await self.crawler.async_send(uncommon_request)
+            except RequestError:
+                self.network_errors += 1
+                return
+            if uncommon_response.status != 405:
+                # If the method is HEAD and its body is empty, this behavior is not abnormal
+                methods_dict.update({method: {"code": uncommon_response.status,
+                                              "status_different": response.status != uncommon_response.status,
+                                              "content_different": False
+                                              if (method == 'HEAD' and uncommon_response.content == "")
+                                              else response.content != uncommon_response.content,
+                                              "response": uncommon_response,
+                                              "request": uncommon_request
+                                              }})
+
+        return methods_dict
 
     async def must_attack(self, request: Request, response: Optional[Response] = None):
         return request.path not in self.excluded_path
@@ -56,24 +112,59 @@ class ModuleMethods(Attack):
         log_verbose(f"[+] {option_request}")
 
         try:
-            response = await self.crawler.async_send(option_request)
+            option_response = await self.crawler.async_send(option_request)
         except RequestError:
             self.network_errors += 1
             return
 
-        if response.is_success or response.is_redirect:
-            methods = response.headers.get("allow", '').upper().split(',')
+        # If options response status code is 2** or 301/2/3/5/7
+        option_exist = (option_response.is_success or option_response.is_redirect)
+        if option_exist:
+            methods = option_response.headers.get("allow", '').upper().split(',')
             methods = {method.strip() for method in methods if method.strip()}
-            interesting_methods = sorted(methods - self.KNOWN_METHODS)
+            log_orange(f"Methods found in the header: {','.join(methods)}")
+        else:
+            # Giving an empty method set to the discover method will
+            # make it try all the methods blindly
+            log_orange("No methods found in the header, blindly try all the methods")
+            methods = {}
+        interesting_methods = await self.discover_methods(request, response, methods)
 
-            if interesting_methods:
-                log_orange("---")
-                log_orange(f"Interesting methods allowed on {page}: {', '.join(interesting_methods)}")
-                await self.add_addition(
-                    category=NAME,
-                    request=option_request,
-                    info=f"Interesting methods allowed on {page}: {', '.join(interesting_methods)}",
-                    wstg=WSTG_CODE,
-                    response=response
-                )
-                log_orange("---")
+        if not interesting_methods:
+            log_orange("No interesting method found")
+            return
+
+        log_orange(interesting_methods)
+        log_orange("---")
+        # methods returned by the OPTIONS method request
+        option_log = f"Interesting methods allowed on {page}: {', '.join(list(interesting_methods.keys()))}"
+        if option_exist:
+            log_orange(option_log)
+            await self.add_addition(
+                category=NAME,
+                request=option_request,
+                info=option_log,
+                wstg=WSTG_CODE,
+                response=option_response
+            )
+        # if a method has a relevant information, it will be listed
+        for method, data in interesting_methods.items():
+            if not (data['status_different'] or data['content_different']):
+                continue
+            logging_string = f"Method {method} returned "
+            differences_str = []
+            if data['status_different']:
+                differences_str.append(f"{data['code']} server code")
+            if data['content_different']:
+                differences_str.append("a body content")
+            logging_string += f"{' and '.join(differences_str)} different from GET method on {page}"
+            log_orange(logging_string)
+            await self.add_addition(
+                category=NAME,
+                request=data['request'],
+                info=logging_string,
+                wstg=WSTG_CODE,
+                response=data['response']
+            )
+        if option_exist or interesting_methods:
+            log_orange("---")
--- a/tests/attack/test_mod_methods.py
+++ b/tests/attack/test_mod_methods.py
@@ -13,43 +13,253 @@ from wapitiCore.attack.mod_methods import ModuleMethods
 
 @pytest.mark.asyncio
 @respx.mock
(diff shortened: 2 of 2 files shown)


wapitiCore/attack/mod_methods.py
```diff
@@ -59,21 +102,46 @@ async def attack(self, request: Request, response: Optional[Response] = None):
...
+                    logging_string = ""
+                    if data['diff_code'] and data['diff_content']:
+                        logging_string = f"Method {method} returned {data['code']} server code "\
+                            "and its body content differs from GET method on {page}"
+                    elif data['diff_code']:
+                        logging_string = f"Method {method} returned {data['code']} server code "\
+                            "which is different from GET method on {page}"
+                    elif data['diff_content']:
+                        logging_string = f"Method {method} returned a body content different"\
+                            " from GET method on {page}"
```
cedricvoinnet, line 137:
```suggestion
                    next if not data['diff_code'] and not data['diff_content']

                    logging_string = f"Method {method} returned "
                    differences_str = []
                    if data['diff_code']:
                        differences_str.append(f"{data['code']} server code")
                    if data['diff_content']:
                        differences_str.append(f"a body content")
                    logging_string += f"{' and '.join(differences_str)} different from GET method on {page}"
```

wapitiCore/attack/mod_methods.py
```diff
@@ -59,21 +102,46 @@ async def attack(self, request: Request, response: Optional[Response] = None):
...
+                    if logging_string:
+                        log_orange(logging_string)
+                        await self.add_addition(
+                            category=NAME,
+                            request=option_request,
```
cedricvoinnet, line 142: It should be the current request rather than the option_request

wapitiCore/attack/mod_methods.py
```diff
@@ -39,6 +40,48 @@ class ModuleMethods(Attack):
...
+            if response.status == 405:
+                not_allowed_methods.add(method)
+            else:
+                # Nested dict, each method has a dictionnary containing
+                # the returned server code, if its different from the
+                # get method, and if its body is different from the
+                # get method
+                response_dict.update({method: {"code": response.status,
+                                               "diff_code": get_response.status != response.status,
+                                               "diff_content": get_response.content != response.content,
```
cedricvoinnet, line 79:
```suggestion
                                               "status_different": get_response.status != response.status,
                                               "content_different": get_response.content != response.content,
```

cedricvoinnet (changes requested):

wapitiCore/attack/mod_methods.py
```diff
@@ -64,16 +106,44 @@ async def attack(self, request: Request, response: Optional[Response] = None):
...
+                            request=Request(
+                                page,
+                                method,
+                                referer=request.referer,
+                                link_depth=request.link_depth
+                            ),
+                            info=logging_string,
+                            wstg=WSTG_CODE,
+                            response=data['resp']
+                        )
```
Slokilla, line 148: You made this part of method way smarter than before, we should refactor this before before you go right into an indentation hell. 

What I recommend is transforming the structure : 

```python
if A
   if B
     if C
```

into : 
```python
if not A: return
if not B: return
if not C: return 
```

wapitiCore/attack/mod_methods.py
```diff
@@ -39,6 +39,48 @@ class ModuleMethods(Attack):
...
+            except RequestError:
+                self.network_errors += 1
+                return
+            if response.status == 405:
+                not_allowed_methods.add(method)
+            else:
+                # Nested dict, each method has a dictionnary containing
+                # the returned server code, if its different from the
+                # get method, and if its body is different from the
+                # get method
```
Slokilla, line 75: Well named variables are explicit enough imho, you can remove the comment.

wapitiCore/attack/mod_methods.py
```diff
@@ -39,6 +39,48 @@ class ModuleMethods(Attack):
...
+            get_response = await self.crawler.async_send(request)
+        except RequestError:
+            self.network_errors += 1
+            return
+        # Testing and comparing the uncommon methods,
+        # not_allowed_methods record all the methods
+        # returning a 405 error which will be removed
+        not_allowed_methods = set()
+        for method in methods:
+            uncommon_request = Request(
```
Slokilla, line 57: is it important to keep specify that the request is uncommon ? 

Naming it _request_ should be clear enough.

wapitiCore/attack/mod_methods.py
```diff
@@ -39,6 +39,48 @@ class ModuleMethods(Attack):
...
     excluded_path = set()
 
+    async def test_compare_methods(self, request: Request, methods: set):
+        response_dict = {}
+        # posting a GET request to compare
+        # with the uncommon methods
+        try:
+            log_verbose(f"[+] {request}")
+            get_response = await self.crawler.async_send(request)
+        except RequestError:
```
Slokilla, line 49: Is it relevant to make a request without verifying that GET is an option ? Let's save a request by verifying that GET exists before calling this method.

wapitiCore/attack/mod_methods.py
```diff
@@ -39,6 +39,48 @@ class ModuleMethods(Attack):
...
+                link_depth=request.link_depth
+            )
+            try:
+                log_verbose(f"[+] {uncommon_request}")
+                response = await self.crawler.async_send(uncommon_request)
+            except RequestError:
+                self.network_errors += 1
+                return
+            if response.status == 405:
+                not_allowed_methods.add(method)
```
Slokilla, line 70: Do you really need to have a second set? Maybe you could remove methods on the go, like

```suggestion
                methods - method
```

If you do it this way, you must modify line 82 to return _sorted(methods)_

wapitiCore/attack/mod_methods.py
```diff
@@ -39,6 +39,48 @@ class ModuleMethods(Attack):
     do_post = True
     excluded_path = set()
 
+    async def test_compare_methods(self, request: Request, methods: set):
```
Slokilla, line 42: You should think about renaming this. It is unclear because :
1 - It's not a test in the way it asserts nothing
2 - What will this function compare ?
3 - Is this only a comparison ? It returns a set of methods, and it is not reflected in the method name.

Slokilla (changes requested):

wapitiCore/attack/mod_methods.py
RMI78: request is already refering to an argument of the function. In fact, it is the OPTIONS request that is passed in argument, I could rename it to free the name "request" but i will have to rename the argument variable into something like "option_request", I guess it will just shift the problem around.

wapitiCore/attack/mod_methods.py
```diff
@@ -56,24 +108,49 @@ async def attack(self, request: Request, response: Optional[Response] = None):
...
-                await self.add_addition(
-                    category=NAME,
-                    request=option_request,
-                    info=f"Interesting methods allowed on {page}: {', '.join(interesting_methods)}",
-                    wstg=WSTG_CODE,
-                    response=response
-                )
-                log_orange("---")
+        if not (option_response.is_success or option_response.is_redirect):
+            return
```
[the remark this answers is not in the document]
Slokilla: If we do this, we should ad a log line to say that OPTIONS has no response, and that we are going to try an hardcoded list of methods.

Slokilla (changes requested):

wapitiCore/attack/mod_methods.py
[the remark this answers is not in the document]
devl00p: @Slokilla I think you mean 405 ?

I think the logic should be changed in the following way:

- if `OPTIONS` returns the `Allow` header, rely solely on it
- if it returns a 405 test each method separately, ignore 403 and 405 responses otherwise compare responses

wdyt?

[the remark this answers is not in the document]
devl00p: Here is a case I saw on a CTF where `OPTIONS` returned a 405 but `PUT` was allowed : https://devl00p.github.io/posts/Solution-du-CTF-Rubbish-de-Wizard-Labs/#claude-rish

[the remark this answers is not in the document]
Slokilla: Actually, I mean anything that differ from 200 because the way that OPTIONS failed isn't that important.

So i would modify a bit your logic to: 
  - if OPTIONS returns the _Allow_ header, rely solely on it
  - if it fails test each method separately [...]

I hope I'm not talking too much nonsense ^^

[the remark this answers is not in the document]
RMI78: If we are heading that way, shouldn't we start considering both 404 and 405 status code, since 404 will indeed fall into the exception ? Moreover, a server can totally customize the way he can answer to any type of request, so I'm confused regarding the type of behavior that can be relevant to us. 

I mean, any unexpected returned headers might be done on purpose or results from an unexpected behavior from the server. 
We should see every other return code similar to 405 and 404 returned by the ``OPTIONS`` method knowing that the method module will try to see differences between all the methods (or at least, the one specified in the ``Allow`` header with the ``OPTIONS`` method) and the get request made by the crawler.

Plus, a ``HEAD`` request is supposed to return the same thing as the ``GET`` request but without the body  (this can also be bypassed with custom headers), however, for now it is still triggered, and I was wondering if creating a special case for that behavior could be a good idea.

What are your thoughts on those points ?

Slokilla (changes requested):

wapitiCore/attack/mod_methods.py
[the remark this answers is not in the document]
devl00p: For the OPTIONS response indeed I think we can rely only on the Allow header whatever the status code.

Then for custom methods, ignore 400/401/403/404/405 responses ? I think the status code is explicit.

Maybe we can also check for false positives using the status code returned by an invalid method ?

```console
$ curl -s -D- -XOPTIONS https://talk.jekyllrb.com/ | head -1
HTTP/2 404 
$ curl -s -D- -XPUT https://talk.jekyllrb.com/ | head -1
HTTP/2 404 
$ curl -s -D- -XLOL https://talk.jekyllrb.com/ | head -1
HTTP/2 405 
$ curl -s -D- -XGET https://talk.jekyllrb.com/ | head -1
HTTP/2 200
```
vs

```console
$ curl -s -D- -XGET https://www.tevora.com/ | head -1
HTTP/2 200 
$ curl -s -D- -XOPTIONS https://www.tevora.com/ | head -1
HTTP/2 200 
$ curl -s -D- -XPUT https://www.tevora.com/ | head -1
HTTP/2 200 
$ curl -s -D- -XLOL https://www.tevora.com/ | head -1
HTTP/2 200
```

wapitiCore/attack/mod_methods.py
```diff
@@ -56,24 +112,54 @@ async def attack(self, request: Request, response: Optional[Response] = None):
...
+            methods = {}
+        interesting_methods = await self.discover_methods(request, response, methods)
+        log_orange(interesting_methods)
+
+        if not interesting_methods:
+            return
+
+        log_orange("---")
+        # methods returned by the OPTIONS method request
+        option_log = f"Interesting methods allowed on {page}: {', '.join(list(interesting_methods.keys()))}"
```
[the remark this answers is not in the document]
RMI78: Well, it still mostly rely on the OPTIONS method to try all the other ones or blindly try all the methods specifically if the OPTIONS method returns nothing so it still makes sens to me.

Slokilla (approved):

PR #397, opened 2023-04-30
Reduce image test size

Slokilla:
This MR features simple operations to reduce the test docker image.

With these changess, we build a 40% lighter image.
First, we regroup all apt install lines, so we can remove cache on the
same layer.

Then we remove pip3 cache, which is absolutely not negligible.


bretfourbe:
Maybe we should do the same for github actions ?

bretfourbe (approved):

Issue #396, opened 2023-04-27
With the current options, we cannot have the "%" char in the login.


Slokilla:
Quoting the wapiti man 

> - -a, --auth-cred
    Set credentials to use for HTTP authentication on the target (see available methods bellow). Given value should be in the form 
    login%password (% is used as a separator)

Let's suppose that my login is "99%OfCoffee", and my password "1%OfBrain". What would happen with the current CLI ? 

I suppose that the used login will be "99", and the used password "OfCoffee1%OfBrain". What a pity !

Can we provide more sophisticated options like "--login" and "--password" to provide both separately. This way, we keep supporting the old syntax, and we add support of "%" in logins.

devl00p:
Good idea. I think  `--user` would be more meaningful that `--login`.

Also we should certainly name them `--auth-user` and `--auth-password` as we already have the `--auth-method` option. That way users whould not be mistaken with other credentials.

By the way the same should be done for `--form-cred`

Slokilla:
I totally agree with your suggestion !

PR #405, opened 2023-05-15
Fix Issue #396

Darkiros:
Fix issue #396.
I added the suggested changes and left the old system but tag as deprecated in thoses files.
I also changed tests files with the new system.


wapitiCore/main/getcookie.py
```diff
@@ -101,14 +107,35 @@ async def getcookie_main(arguments):
...
     )
 
+    # This option is deprecated
+    # Should be removed in a future version
     parser.add_argument(
         "-a", "--auth-cred",
         dest="http_credentials",
+        action="append",
         default=argparse.SUPPRESS,
         help="Set HTTP authentication credentials",
```
cedricvoinnet, line 117: We should indicate in the help here that this option is deprecated.

wapitiCore/parsers/commandline.py
```diff
@@ -112,29 +112,71 @@ def parse_args():
...
     )
 
+    # This option is deprecated
+    # Should be removed in a future version
     parser.add_argument(
         "-a", "--auth-cred",
         dest="http_credentials",
+        action="append",
         default=argparse.SUPPRESS,
         help="Set HTTP authentication credentials",
```
cedricvoinnet, line 122: We should indicate in the help here that this option is deprecated.

wapitiCore/parsers/commandline.py
```diff
@@ -112,29 +112,71 @@ def parse_args():
...
     )
 
+    # This option is deprecated
+    # Should be removed in a future version
     parser.add_argument(
         "--form-cred",
         dest="form_credentials",
+        action="append",
         default=argparse.SUPPRESS,
         help="Set login form credentials",
```
cedricvoinnet, line 158: We should indicate in the help here that this option is deprecated.

cedricvoinnet (approved):

bretfourbe:
Hi @devl00p ! What do you think about this ? Is it ok ?

wapitiCore/parsers/commandline.py
```diff
@@ -112,27 +112,69 @@ def parse_args():
         type=float
     )
 
+    # This option is deprecated
+    # Should be removed in a future version
     parser.add_argument(
         "-a", "--auth-cred",
         dest="http_credentials",
+        action="append",
```
devl00p, line 120: It looks good overall but I think you should avoid the "append" action and use the default "store" action otherwise it may lead to unexpected behavior, for example if someone uses `--auth-user` and `--auth_cred`.

It seems more stable imho to use two variables explicitely for user and password.

devl00p (changes requested):

bretfourbe (approved):

devl00p (approved):

Issue #407, opened 2023-05-24
Verify XXE payload through uploaded files


devl00p:
I recently bumped into that CTF : https://vulnhub.com/entry/minu-v2,333/

There is an upload form accepting SVG files only and the form is vulnerable to XXE.

This is a special case because the vulnerability is triggered by an action once the file is uploaded.

Wapiti didn't detect the vulnerability while it should (`attack_upload` in `mod_xxe` was made for this)

We should check what is hapenning here to understand if it is a bug or if we can improve the module.

devl00p:
Turns out the form was only accepting files with .svg extensions.

Did an MR to test upload with both xml and svg extensions. Made it easy to add another extension if we need it (docx ?)

PR #409, opened 2023-05-25
Renamed FileMutator to XXEUploadMutator …

devl00p:
as it is specific to XXE attacks.

Added a list of suffix to test when uploading file for XEE


devl00p:
fix #407

PR #411, opened 2023-06-05
Add tests and module to test https redirections

bretfourbe:
Add test files and new module to test https redirections.
This module tries to replay requests over HTTP.
Add new definition file.


polyedre (approved):

devl00p:
I started looking at this module

Should we raise an informational message if the [Strict-Transport-Security](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security) is missing in the http response ?

devl00p:
Also what about making a first check to see if the website is available on the http port ? If it isn't listening then stop the module

bretfourbe:
> I started looking at this module
> 
> Should we raise an informational message if the [Strict-Transport-Security](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security) is missing in the http response ?

Hi @devl00p, this kind of issue is already handled in `http_headers` module, and regarding https://github.com/wapiti-scanner/wapiti/issues/282 i don't think it is relevant for HTTP

bretfourbe:
> Also what about making a first check to see if the website is available on the http port ? If it isn't listening then stop the module

Yes indeed, i will add this :)

wapitiCore/attack/mod_https_redirect.py
```diff
@@ -0,0 +1,69 @@
...
+    async def attack(self, request: Request, response: Optional[Response] = None):
+
+        url = urlparse(request.url)
+        # ensure targeting http url
+        http_url = urlunparse(url._replace(scheme='http'))
+        http_request = Request(
+            http_url, request.method,request.get_params,
+            request.post_params, request.file_params, request.encoding,
+            request.enctype, request.referer, request.link_depth
+        )
```
Slokilla, line 31: Would it work with unpacking ? 

`http_request = Request(http_url, **request)`

Slokilla (commented):
You should modify the project's README.md to add this in the "Supported attacks" section

Slokilla (approved):

bretfourbe:
@devl00p 

> > Also what about making a first check to see if the website is available on the http port ? If it isn't listening then stop the module
> 
> Yes indeed, i will add this :)

I added the check in the `must_attack` method

wapitiCore/attack/mod_https_redirect.py
```diff
@@ -0,0 +1,91 @@
...
+
+from wapitiCore.attack.attack import Attack
+from wapitiCore.definitions.https_redirect import NAME, WSTG_CODE
+from wapitiCore.definitions.internal_error import WSTG_CODE as INTERNAL_ERROR_WSTG_CODE
+from wapitiCore.language.vulnerability import Messages
+from wapitiCore.net import Request
+from wapitiCore.net.response import Response
+from wapitiCore.main.log import log_red, log_orange
+
+def test_port(address: str, dest_port: int, timeout: float = None) -> bool:
```
devl00p, line 15: You should use the `@lru_cache` decorator to prevent connection tests from being repeated

wapitiCore/attack/mod_https_redirect.py
```diff
@@ -0,0 +1,91 @@
...
+    name = "https_redirect"
+    MSG_VULN_NO_REDIRECT = "No HTTPS redirection"
+    MSG_VULN_REDIRECT = "Redirected to HTTP"
+
+    async def must_attack(self, request: Request, response: Optional[Response] = None):
+        if self.finished:
+            return False
+
+        url = urlparse(request.url)
+        if not test_port(url.hostname, url.port or 80, self.crawler.timeout):
```
devl00p, line 38: There is something to fix about the tested port, here in `must_attack` or/and in `attack`.

If the URL received as parameter is for example http://perdu.com:8888/ then the port being tested would be 8888 and it seems ok but does it makes sense as Wapiti tested the URL while crawling ?

However if the URL is http**s**://perdu.com:8843/ then the tested port would be 8843 which is incorrect (it should be 80) while it makes more sense to test the port.

Also do we want to test all URLs found by the crawler (http or https) or just consider https urls ? It makes more sense to consider only https URLs as someone scanning an http-only website on purpose certainly already knows the risks while someone feeding Wapiti with an https URL might be interested in downgrade attacks.

Wdyt ?

bretfourbe: Here is the new behavior : 

- if http url provided, it will stop the module after the first request and if there is no redirection to https, it will also raise a vuln
- if specific port (8080 or 8443 for instance) is provided, it will stop the module after the first request, if it is https we will probably get a 400 error, if it is http it will raise a vuln only if there is no redirection to https
- in other cases, it will test all the requests

fwininger (approved):

bretfourbe:
@devl00p is it good for you ?

wapitiCore/attack/mod_https_redirect.py
```diff
@@ -0,0 +1,111 @@
...
+        )
+
+        try:
+            http_response: Response = await self.crawler.async_send(http_request, follow_redirects=False)
+        except RequestError:
+            self.network_errors += 1
+            return
+
+        if http_response.is_success:
+            log_red(http_response.url)
```
devl00p, line 80: You should make the log message more detailed, for example "URL ... doesn't redirect to https", otherwise LGTM

devl00p (approved):

bretfourbe:
@devl00p should i merge this one even if the codecov job failed ?

PR #413, opened 2023-06-16
Fix empty and multiple values headers

RMI78:
@devl00p,  @bretfourbe, @fwininger 
While testing the wapp module that use wappalyzer, I realized the module mess up with some headers configurations

Firstly, when the regex is empty in our JSON file, we replace it with its attribute as you can see [here](https://github.com/wapiti-scanner/wapiti/blob/9c91017c016ec82df9f0ebf78cd2040a34b1c93b/wapitiCore/wappalyzer/wappalyzer.py#LL131C1-L136C39)
But then we compare this value (the copied attribute) with its value from the response instead of checking if it is simply here, which will not raise it.

The second bug I found was with headers that have multiple values for the same attribute. Let's admit a response has 2 headers `Server` (which will most likely never happens for this attribute but whatever):
```Text
Server : this 
Server : that
```
The code use to concatenate each value from attributes in a dictionary with a single string that way:
```JSON
{
   "Server": "this, that"
}
```
Which will make some regexes fail: admitting we can detect a technology thanks to its server attribute "that" by its regex, which is "^that$". It will not work as the regex is expecting the string to start and end with this word. As this behavior comes from HTTPX, you can find [more information here](https://www.python-httpx.org/quickstart/#response-headers).

Instead of using a split method (because we can't take for granted the fact that all the attributes will never have a " ," as values inside of them) I safely reparsed the dictionary from the httpx.Header object, so now we have a dictionary that looks more like this:
```JSON
{
   "Server": ["this", "that"]
}
```
And the regex search for each element.

I also found a third bug in which the keys provided by the JSON database are set in lowercases, meanwhile the keys of the dictionary made by the httpx.Response for the cookies are in uppercase. I lowered down all the keys, so it can detect all the technos
 
Finally, I've written some more unit tests for those cases to check that everything runs smoothly.

EDIT 1:  I found out that 
```Python
 expected_result = [
            '{"name": "Astra Widgets", "versions": ["1.5.4"], "categories": ["WordPress plugins", "Widgets"], "groups": ["Add-ons", "Other"]}',
            '{"name": "PHP", "versions": [], "categories": ["Programming languages"], "groups": ["Web development"]}'

        ]
for arg in persister.add_payload.call_args_list:
            assert arg[1]["info"] in expected_result
```
Is a bad way to check the units tests since we can put technos in expected values that couldn't be in the persister, so I patched them using sets instead which guarantees the existence and uniqueness of each expected strings: 
```Python
 expected_result = [
            '{"name": "Astra Widgets", "versions": ["1.5.4"], "categories": ["WordPress plugins", "Widgets"], "groups": ["Add-ons", "Other"]}',
            '{"name": "PHP", "versions": [], "categories": ["Programming languages"], "groups": ["Web development"]}'

        ]
assert set([arg[1]['info'] for arg in persister.add_payload.call_args_list]) == expected_result
```
It turned out that there were more bugs involved than simply the headers, I'll keep working on them even tho the name of the MR is loosing its meaning as this patch is growing bigger

EDIT 2:
In the unit test file, for the DOM, a techno (Joomla) implied by another one (Sellacious) was missing, so I added it and some attributes were set to id whereas they have to be set to class according to the JSON.
I encountered the same issue with the meta tags that I've experienced with the headers tags (the fact that as they are stored in a dictionary, every tag with the same name will be squashed into one regardless of their content). This made me add a new method in the HTML parser to allow a list of tuples ``(name, content)`` so they are all unique (and use the same format as multi_items() from httpx.Headers), and I reparsed them as a dictionary of lists in the Wappalyzer class just like the headers

All the unit tests are now green, and I haven't found anymore bugs so far

### DIFF
--- a/tests/attack/test_mod_wapp.py
+++ b/tests/attack/test_mod_wapp.py
@@ -154,8 +154,8 @@ async def test_dom_detection():
             <h2>Pas de panique, on va vous aider</h2> \
             <strong><pre>    * <----- vous &ecirc;tes ici</pre></strong> \
             <input type=\"hidden\" name=\"_glpi_csrf_token\" value=\"b6db36a8c9fd4f3f5d244faa76247688\">\
-            <p id=\"mod-sellacious-cart\">test text</p> \
-            <p id=\"sm-page-footer-copyright\">SmugMug</p> \
+            <p class=\"mod-sellacious-cart\">test text</p> \
+            <p class=\"sm-page-footer-copyright\">SmugMug</p> \
             <img src=\"www.afi-b.com\" /> \
             <a href=\"/cart\">test</a> \
             </body></html>"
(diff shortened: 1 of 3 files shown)


RMI78:
As this MR modifies the code a lot, it requires being split into multiple smaller MR with further comments if needed.

Slokilla:
As you splitted this merge request, I think you can close it. 

https://github.com/wapiti-scanner/wapiti/pull/434
https://github.com/wapiti-scanner/wapiti/pull/441

PR #417, opened 2023-07-02
Payloads refactoring

devl00p:
This PR brings several things:

#### Flags 

The `Flags` class wasn't very handy : it mainly integrated information about the payload but it wasn't used everywhere. I chose to mix the Flags and payload in a new PayloadInfo class.

Another issue with `Flags` was that it wasn't dynamic so adding an information to a payload required more code.

#### INI parser

Now the PayloadInfo class is generated dynamically out of INI entries using `dataclasses.make_dataclass`.

For this to work I'm reusing the `[DEFAULT]` entry of INI files. There all possible keys for the file should be declared with a default (placeholder) value.

I added the support of the `None` value.

The main benefit of the new bahavior is that adding a new information tied to a payload only require adding two lines to the INI file (one for the `DEFAULT` section, one for the payload section)

For example we may want to add a description for each payload and put that description in the report.

The parser is more agnostic than previously. You can add handlers that are callback that will be applied to the payload. As many callbacks as you want for each section key.

#### TXT parser

Similar to the INI parser, the TXT parser can be given callbacks that will be applied to each line of txt file.

#### Parameter

Another information that was generated by the Mutator is the name of the injected parameter.

I chose to change that for a new `Parameter` class that has a `name` and a `situation` telling where the injection occurs.

It would certainly help to inject data in more places in the future (path, JSON data or XML, etc)

### Fix

I noticed two issus in `mod_permamentxss`. First, with data obtain through the `tried_xss` directory, the mutated request used to test the injection without payload was used as the reference. It made the module crash because the persister tried to get the `path_id` for that request while it is not saved to the db.

I fixed that by giving the original request in `tried_xss`.

I also changed the check made on the content-type as it was checking the page used to send the payload but not the page where the XSS is rendered (`tried_xss` is used for second-order XSS injection).

I also added a first unit test for the  `mod_permamentxss`  module.


devl00p:
Also remove automatic use of payload loading when PAYLOADS_FILE is declared

PR #422, opened 2023-07-10
Fix #394: Network issues and crawlers improvements

### LINKED ISSUE #394: Many network issues identified during scans
Hi @devl00p !
I observed that many network issues are raised during scans, on several modules (ssrf, file, buster, crlf, log4shell).
I changed the code in a module (buster in this case) to get the stack traces below.
On HTTP target (testphp.vulweb.com) :
image
image
On HTTPS target : 
image
image
Do you have any idea ?
Thanks !

Darkiros:
Fix issue #394.

* I fix the network issues with a new configuration in `crawler.py` : 
```python
self._timeout = httpx.Timeout(timeout, read=None)
```

I did many tests to decide how to do. And here is some element to compare why None is better : 

Here with 10s in read timeout
image

Here with None in read timeout
image

The crawler is faster and crawl more URL than with 10s timeout. There is no chance to be stuck with the response delay because connect, pool and write timer are set to default or -t option.

* I also set a new param in different send request to allow the overwriting of the new crawler configuration for some modules like `timesql` and change the `read=None` to the -t or the default value.

* I change the timeout in headless crawler to fit with the -t option instead of constant value.

* I make changes with the new behaviour in test_network_issues modules.

### DIFF
--- a/wapitiCore/attack/attack.py
+++ b/wapitiCore/attack/attack.py
@@ -310,9 +310,9 @@ class Attack:
             skip=self.options.get("skipped_parameters")
         )
 
-    async def does_timeout(self, request):
+    async def does_timeout(self, request, timeout: float = None):
         try:
-            await self.crawler.async_send(request)
+            await self.crawler.async_send(request, timeout=timeout)
         except ReadTimeout:
             return True
         except RequestError:
--- a/wapitiCore/net/crawler.py
+++ b/wapitiCore/net/crawler.py
@@ -98,7 +98,7 @@ class AsyncCrawler:
     ):
         self._base_request = base_request
         self._client = client
-        self._timeout = timeout
+        self._timeout = httpx.Timeout(timeout, read=None)
 
         self.is_logged_in = False
         self.auth_url: str = self._base_request.url
@@ -217,7 +217,8 @@ class AsyncCrawler:
             resource: web.Request,
             follow_redirects: bool = False,
             headers: dict = None,
-            stream: bool = False
+            stream: bool = False,
+            timeout: float = None,
     ) -> Response:
         """Fetch the given url, returns a Response object on success, None otherwise.
         If None is returned, the error code can be obtained using the error_code property.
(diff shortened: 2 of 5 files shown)


wapitiCore/net/crawler.py
```diff
@@ -231,7 +232,11 @@ async def async_get(
         @type stream: bool
         @rtype: Response
         """
-        request = self._client.build_request("GET", resource.url, headers=headers, timeout=self.timeout)
+        if timeout is None:
+            request = self._client.build_request("GET", resource.url, headers=headers, timeout=self.timeout)
+        else:
+            timeout = httpx.Timeout(timeout)
+            request = self._client.build_request("GET", resource.url, headers=headers, timeout=timeout)
```
Slokilla, line 239: a more compact way is 
```suggestion
        timeout = self.timeout if timeout is none else httpx.Timeout(timeout)
        request = self._client.build_request("GET", resource.url, headers=headers, timeout=timeout)

```

wapitiCore/net/crawler.py
```diff
@@ -289,16 +295,29 @@ async def async_post(
...
+            request = self._client.build_request(
+                "POST",
+                form.path,
+                params=form.get_params,
+                data=post_params,  # httpx expects a dict, hope to see more types soon
+                content=content,
+                files=file_params or None,
+                headers=form_headers,
+                timeout=timeout
+            )
```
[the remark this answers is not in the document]
Slokilla: Idem for below usage

Slokilla (changes requested):

wapitiCore/attack/attack.py
```diff
@@ -310,9 +310,9 @@ def get_mutator(self):
             skip=self.options.get("skipped_parameters")
         )
 
-    async def does_timeout(self, request):
+    async def does_timeout(self, request, timeout = None):
```
devl00p, line 313: can you add type annotation please ?

PR #436, opened 2023-07-24
Add module spring4shell

Darkiros:
Based on @bretfourbe 's work. #275 

I changed the detection method to be less aggressive and also added unit test.

### DIFF
--- a/wapitiCore/attack/mod_spring4shell.py
+++ b/wapitiCore/attack/mod_spring4shell.py
@@ -1,0 +1,91 @@
+import uuid
+
+from typing import Optional
+from httpx import RequestError
+from wapitiCore.attack.attack import Attack
+from wapitiCore.definitions.spring4shell import NAME, WSTG_CODE
+from wapitiCore.main.log import log_red, logging
+from wapitiCore.net import Request, Response
+
+
+class ModuleSpring4Shell(Attack):
+    """
+    Detect the Spring4Shell vulnerability
+    """
+
+
+    async def _default_request(self, request_url: str, method: str) -> int:
+        request = Request(
+            path=request_url,
+            method=method,
+        )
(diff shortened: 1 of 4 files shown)


devl00p (approved):

wapitiCore/attack/mod_spring4shell.py
```diff
@@ -0,0 +1,103 @@
...
+            return
+
+
+    async def attack(self, request: Request, response: Optional[Response] = None):
+        await self._attack_spring4shell_url(request)
+
+
+    async def _check_spring4shell(self, method: str, request: Request, payload: str) -> bool:
+        key_payload = payload.split("=")[0]
+        value_payload = payload.split("=")[1]
```
RMI78, line 51: No need to use indexes, Python unpacking variable will do the job
```Python
key_payload, value_payload = payload.split("=")
```

wapitiCore/attack/mod_spring4shell.py
```diff
@@ -0,0 +1,103 @@
...
+            )
+        else:
+            params = request.post_params
+            params.append([key_payload, value_payload])
+
+            malicious_request = Request(
+                path=request.url,
+                method=method,
+                post_params=params,
+            )
```
RMI78, line 70: There's a similar pattern in your 2 conditions, here's a shortened version, the boolean variable may need to be shortened tho 
```Python
get_True_post_False = (method=='GET')
params = requests.get_params if get_True_post_False else requests.post_params
params.append([key_payload, value_payload])
            malicious_request = Request(
                path=request.url,
                method=method,
                get_params=params if get_True_post_False else None,
                post_param=params if not get_True_post_False else None
            )
```

bretfourbe (approved):

PR #437, opened 2023-07-24
Add module for Unrestricted uploads vulnerability

devl00p:
* adds module to test for unrestricted file uploads by sending PHP file with several extensions
* fix how excluded urls were used (was some old code left after refactoring)
* pass excluded urls through attack options so the module doesn't attempt to fetch those urls

### DIFF
--- a/wapitiCore/net/scope.py
+++ b/wapitiCore/net/scope.py
@@ -16,6 +16,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+import re
 from typing import Iterable, Union, Set
 from urllib.parse import urlparse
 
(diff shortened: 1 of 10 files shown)


wapitiCore/net/scope.py
```diff
@@ -71,3 +72,21 @@ def check(self, resource: Union[Request, str]) -> bool:
...
+
+def wildcard_translate(pattern: str) -> re.Pattern:
+    """Translate a wildcard PATTERN to a regular expression object.
+
+    This is largely inspired by fnmatch.translate.
+    """
+
+    i, length = 0, len(pattern)
+    res = ''
+    while i < length:
```
Slokilla, line 85: i think `for char in pattern:` should do the job, no ?

wapitiCore/net/scope.py
```diff
@@ -71,3 +72,21 @@ def check(self, resource: Union[Request, str]) -> bool:
...
+        char = pattern[i]
+        i += 1
+        if char == '*':
+            res += r'.*'
+        else:
+            res += re.escape(char)
```
Slokilla, line 91:
```suggestion
        res += r'.*' if char == '*' else re.escape(char)
```

wapitiCore/net/scope.py
devl00p: Indeed, that code deserved some rewriting :) Added some tests too

PR #438, opened 2023-07-25
adding base integration test system

RMI78:
MR for the integration test system. This is the base system with the documentation, tests will come once this MR will be merged.

As of today, I can and will provide integration tests for the following modules:
- backup 
- wp_enum 
- drupal_enum
- brute_login_form 
- cookieflags 
- crlf 
- csp 
- exec 
- file 
- htaccess 
- http_headers 
- xss 
- shellshock 
- methods 
- timesql 
- wapp
- permanentxss
- log4shell
- redirect
- ssrf

As well as a test for the crawler and its different way to handle redirections, plus a test for every authenticating methods it does. 

You can look up in the documentation to see how it works

PR #440, opened 2023-07-26
Test wapp module

RMI78:
⚠️  Depends on #438 

This is a first test for the integration test system, it is testing the wapp module. That way, you can see how it works.


tests/integration/run.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-# List of modules to be tested
-TESTS=""
+# List of the currently working tests
```
Slokilla, line 3: rebase issue ?

RMI78: clearly

RMI78: shouldn't be a problem since the TESTS variable is replaced

Slokilla (approved):

bretfourbe (approved):

fwininger (approved):

PR #442, opened 2023-07-27
Update CSP module

OussamaBeng:
adding more information about the CSP directives and the unsafe values to avoid

### DIFF
--- a/wapitiCore/attack/mod_csp.py
+++ b/wapitiCore/attack/mod_csp.py
@@ -21,16 +21,68 @@ from httpx import RequestError
 from wapitiCore.attack.attack import Attack
 from wapitiCore.net import Request
 from wapitiCore.net.response import Response
-from wapitiCore.net.csp_utils import csp_header_to_dict, CSP_CHECK_LISTS, check_policy_values
+from wapitiCore.net.csp_utils import csp_header_to_dict, CSP_CHECK_LISTS
 from wapitiCore.definitions.csp import NAME, WSTG_CODE
 from wapitiCore.main.log import log_red
 
 MSG_NO_CSP = "CSP is not set"
-MSG_CSP_MISSING = "CSP attribute \"{0}\" is missing"
-MSG_CSP_UNSAFE = "CSP \"{0}\" value is not safe"
 
+INFO_UNSAFE_INLINE = "\"unsafe-inline\" in \"{0}\" directive allows the execution of unsafe in-page scripts and event\
+ handlers."
+INFO_UNSAFE_EVAL = "\"unsafe-eval\" in \"{0}\" directive allows the execution of code injected into DOM APIs such as\
+ eval()."
+INFO_DATA_HTTP_HTTPS = "value \"{0}\" URI in \"{1}\" allows the execution of unsafe scripts."
+INFO_ALLOW_ALL = "\"{0}\" directive should not allow \"*\" as source"
+INFO_UNSAFE_OBJECT_SRC = "unsafe values \"{0}\" other then \"none\" identified in \"object-src\""
+INFO_UNSAFE_BASE_URI = "unsafe values \"{0}\" other then \"none\" and \"self\" identified in \"base-uri\""
+INFO_UNDEFINED_DIRECTIVE = "directive \"{0}\" is not defined"
 
 # This module check the basics recommendations of CSP
+def check_policy(policy_name, csp_dict):
+    """
+    This function return the unsafe values for each directive of the tested CSP
+    """
+    info = ""
+
+    if policy_name not in csp_dict and "default-src" not in csp_dict:
+        log_red(INFO_UNDEFINED_DIRECTIVE.format(policy_name))
+        info += INFO_UNDEFINED_DIRECTIVE.format(policy_name) + "\n"
+
+    # The HTTP CSP "default-src" directive serves as a fallback for the other CSP fetch directives.
+    # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/default-src
+    policy_values = csp_dict.get(policy_name) or csp_dict["default-src"]
+    info = ""
+    # If the tested element is default-src or script-src, we must ensure that none of this unsafe values are present
+    if policy_name in ["default-src", "script-src"]:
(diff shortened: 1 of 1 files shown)


wapitiCore/attack/mod_csp.py
```diff
@@ -21,16 +21,65 @@
...
+INFO_UNSAFE_BASE_URI = "unsafe values \"{0}\" other then \"none\" and \"self\" identified in \"base-uri\""
 
 # This module check the basics recommendations of CSP
+def check_policy(policy_name, csp_dict):
+    """
+    This function return the unsafe values for each directive of the tested CSP
+    """
+
+    if policy_name not in csp_dict and "default-src" not in csp_dict:
+        return 0
```
RMI78, line 46: Return 0 when expecting a string isn't consistent

wapitiCore/attack/mod_csp.py
```diff
@@ -65,21 +114,14 @@ async def attack(self, request: Request, response: Optional[Response] = None):
...
-
-                    log_red(info)
-                    await self.add_vuln_low(
-                        category=NAME,
-                        request=request_to_root,
-                        info=info,
-                        wstg=WSTG_CODE,
-                        response=response
-                    )
+                info += check_policy(policy_name, csp_dict)
```
RMI78, line 119: Currently, we check if the most important rules are well set
Shouldn't we also check for mistyped/non-existant CSP by being more strict ? (notify the user of weird CSP like someone who would have typed ``deflaut-src``)

wapitiCore/attack/mod_csp.py
```diff
@@ -21,16 +21,65 @@
...
+    policy_values = csp_dict.get(policy_name) or csp_dict["default-src"]
+    info = ""
+    # If the tested element is default-src or script-src, we must ensure that none of this unsafe values are present
+    if policy_name in ["default-src", "script-src"]:
+        for unsafe_value in CSP_CHECK_LISTS[policy_name]:
+            if unsafe_value in policy_values:
+                if unsafe_value == "unsafe-inline":
+                    log_red(INFO_UNSAFE_INLINE.format(policy_name))
+                    info += INFO_UNSAFE_INLINE.format(policy_name) + "\n"
+                if unsafe_value in ("data:", "http:", "https:"):
```
devl00p, line 59: I think the blocks after the first `if` should be `elif` so Python doesn't lose time evaluating the expresions

wapitiCore/attack/mod_csp.py
OussamaBeng: At present, we only check four directives, and it is difficult to check this case out.
The current version of the module will state that "default-src" is not in the CSP for "deflaut-src" incorrectly typed, for example.
but adding this verification later could be interesting

wapitiCore/attack/mod_csp.py
```diff
@@ -65,21 +117,14 @@ async def attack(self, request: Request, response: Optional[Response] = None):
...
-                    )
+                info += check_policy(policy_name, csp_dict)
+
+            await self.add_vuln_low(
+                category=NAME,
+                request=request,
+                info=info,
+                wstg=WSTG_CODE,
+                response=response
+            )
```
Slokilla, line 130: Do we always add this vuln ? Even if all the CSP are correctly configured ?

wapitiCore/attack/mod_csp.py
```diff
@@ -21,16 +21,68 @@
...
+                elif unsafe_value == "*":
+                    log_red(INFO_ALLOW_ALL.format(policy_name))
+                    info += INFO_ALLOW_ALL.format(policy_name) + "\n"
+                elif unsafe_value == "unsafe-eval":
+                    log_red(INFO_UNSAFE_EVAL.format(policy_name))
+                    info += INFO_UNSAFE_EVAL.format(policy_name) + "\n"
+
+    # If the tested element is none of the previous list, we must ensure that one of this safe values is present
+    else:
+        for safe_value in CSP_CHECK_LISTS[policy_name]:
```
devl00p, line 74: That loop is wrong as it doesn't do the action described in comment.

For example when tested on github.com it displays 

`unsafe values "['self']" other then "none" and "self" identified in "base-uri"`

while github has the policy `base-uri 'self'`

Instead the loop should iterate over `policy_values` THEN check if the value is inside `CSP_CHECK_LISTS[policy_name]`

However on case like  `base-uri yolo 'self'` it will still display something while `self` is in the policy.

I think the best way would be to use sets like this:

```python
if not set(policy_values) & set(CSP_CHECK_LISTS[policy_name]):
  // log policy values
```

Also `policy_values` should be translated to a string joined by commas when displayed.

wapitiCore/attack/mod_csp.py
devl00p: Indeed, looks like an indentation issue

wapitiCore/attack/mod_csp.py
```diff
@@ -21,16 +21,68 @@
...
 # This module check the basics recommendations of CSP
+def check_policy(policy_name, csp_dict):
+    """
+    This function return the unsafe values for each directive of the tested CSP
+    """
+    info = ""
+
+    if policy_name not in csp_dict and "default-src" not in csp_dict:
+        log_red(INFO_UNDEFINED_DIRECTIVE.format(policy_name))
+        info += INFO_UNDEFINED_DIRECTIVE.format(policy_name) + "\n"
```
devl00p, line 49: Beware, you are setting `info` to an empty string just bellow

wapitiCore/attack/mod_csp.py
```diff
@@ -65,21 +117,14 @@ async def attack(self, request: Request, response: Optional[Response] = None):
             )
         else:
             csp_dict = csp_header_to_dict(response.headers["Content-Security-Policy"])
-
+            info = ""
             for policy_name in CSP_CHECK_LISTS:
```
devl00p, line 121: It is convenient to have all policy names to check in `CSP_CHECK_LISTS` at this line, however it is very misleading that some keys correspond to safe values and the other to unsafe values.

It would be nice to split `CSP_CHECK_LISTS` in two dictionaries to differentiate the cases and prevent an error to come.

CSP is not easy to get so the more explicit the code, the lower the risk of introducing bugs

PR #445, opened 2023-07-31
Test ssrf module

RMI78:
This MR is for inserting integration tests scenario for the ssrf module.
Since its report contains data that can vary but is still needed, the check§.sh calls an external python script for a better analysis.

As it is also needed to check ssrf vulns, an endpoint has been added


bretfourbe:
@RMI78 can you rebase ?

bretfourbe (approved):

PR #452, opened 2023-08-02
Tests permanentxss module

RMI78:
integration tests for the permanentxss module


RMI78:
This MR requires more work as permanentxss module use random values in its payloads. A smart_checker should be use such as the ones used with log4shell or ssrf modules

bretfourbe (approved):

PR #475, opened 2023-08-07
distributed docker-compose file with adapted run.sh

RMI78:
This MR simplifies the integration testing system by splitting the docker-compose file. It should optimize things a lot since it will not launch all the services at once when only some tests are required. The build time will be decreased, and some tests can be fully disabled if they break (without breaking the hole system), and we don't care or if we simply want to disable them for reason X.

Building new integration tests will, by consequence, be faster as you can only build your targets

The file run.sh also had a full refactoring, so understanding the changes might not be easy


RMI78:
Still a work in progress, too much output, some working in progress with docker compose build and some repeated code can be simplified

bretfourbe (approved):

tests/integration/run.sh
```diff
@@ -63,7 +57,29 @@ if [[ -v args["--help"] ]]; then
...
+if [[ -z "${TESTS}" ]]; then
+    TESTS=$(find . -maxdepth 1 -type d -name "test_*" -printf "%P ")
+fi
+readarray -d ' ' -t TESTS_ARRAY <<< "${TESTS}"
+export TESTS
+
+# Building the parameters for docker
+for test in "${TESTS_ARRAY[@]}"; do
+    # removing unwanted newlines because last item has one (why ??)
+    DOCKER_COMPOSE_CONFIG_ARGUMENT+="-f ./${test//$'\n'/}/docker-compose.setup.yml "
```
Slokilla, line 72: c'est voulu l'espace finale ?

tests/integration/run.sh
```diff
@@ -77,33 +93,31 @@ fi
...
 fi
 
 # Start the tests
-if [[ -v args["--debug-containers"] ]]; then
-    docker compose  --progress quiet -f docker-compose.setup.yml up --abort-on-container-exit
-else
-    echo "waiting for healthchecks to start Wapiti"
-    docker compose  --progress quiet -f docker-compose.setup.yml up -d
+echo "waiting for healthchecks to start Wapiti"
+docker compose --progress quiet --project-directory ./ -f .docker-compose.final.yml ${DOCKER_COMPOSE_UP_ARGUMENT}
```
Slokilla, line 109: pour des docker compose up argument, c'est pas pratique de mettre des espaces dans les chaines. On a pas un magie blanche en bash pour faire un [arg1, arg2].join(' ') ?

Slokilla (changes requested):

tests/integration/run.sh
RMI78: L'idée serait de stocker les arguments dans un tableau plutôt que dans une chaine ?

PR #485, opened 2023-08-16
Improving https_redirection module

OussamaBeng:
Adding more verification to the module

### DIFF
--- a/wapitiCore/attack/mod_https_redirect.py
+++ b/wapitiCore/attack/mod_https_redirect.py
@@ -27,6 +27,19 @@ def test_port(address: str, dest_port: int, timeout: float = None) -> bool:
         return False
 
 
+async def check_http_redirection(self, request: Request):
+    try:
+        response: Response = await self.crawler.async_send(request, follow_redirects=False)
+    except RequestError:
+        self.network_errors += 1
+        return
+    if response.status in (301, 302):
+        if response.headers.get('Location', '').startswith('https://'):
+            return True
+
+    return False
+
+
 # This module check whether HTTP requests are redirected to HTTPS or not
 class ModuleHttpsRedirect(Attack):
     """Check for HTTPS redirections."""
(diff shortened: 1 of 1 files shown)


wapitiCore/attack/mod_https_redirect.py
```diff
@@ -27,6 +27,19 @@ def test_port(address: str, dest_port: int, timeout: float = None) -> bool:
         return False
 
 
+async def check_http_redirection(self, request: Request):
+    try:
+        response: Response = await self.crawler.async_send(request, follow_redirects=False)
+    except RequestError:
+        self.network_errors += 1
+        return
+    if response.status in (301, 302):
```
devl00p, line 36: You can simplify this by using attributes `is_redirect` and `redirection_url` on the response object

wapitiCore/attack/mod_https_redirect.py
```diff
@@ -50,12 +63,20 @@ async def attack(self, request: Request, response: Optional[Response] = None):
         url = urlparse(request.url)
 
         is_http = False
+        http_request = Request(
+            request.url, request.method, request.get_params,
+            request.post_params, request.file_params, request.encoding,
+            request.enctype, request.referer, request.link_depth
+        )
         if url.scheme == "http":
```
devl00p, line 71: I think a `return` instructions is missing at the end of the `url.scheme == "http"` block.

Without it the same check is done twice

wapitiCore/attack/mod_https_redirect.py
```diff
@@ -50,12 +63,20 @@ async def attack(self, request: Request, response: Optional[Response] = None):
         url = urlparse(request.url)
 
         is_http = False
+        http_request = Request(
```
devl00p, line 66: You should be able to save a request by using the `response` parameter of the `attack()` function and directly checking its headers

PR #496, opened 2023-09-08
Add subdomain scope option and unit test

Darkiros:
Add subdomain for scope option and unit test

### DIFF
--- a/tests/web/test_scope.py
+++ b/tests/web/test_scope.py
@@ -38,7 +38,11 @@ async def test_scopes():
         "http://perdu.com/subdir/page.html?k=v",
         "http://perdu.com/subdir/page.html",
         "http://lost.com/lost.html",
-        "http://external.tld/external.html"
+        "http://external.tld/external.html",
+        "https://subdomain.perdu.com/",
+        "http://subdomain.perdu.com/",
+        "http://subdomain.perdu.com/page.html",
+        "http://subdomain.perdu.com/subdir/subdirpage.html",
     }
 
     scope = Scope(Request("http://perdu.com/subdir/"), "folder")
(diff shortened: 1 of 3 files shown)


tarraschk:
@devl00p petit ping gentil si tu peux regarder nos travaux ici :)

On aimerait avoir ton avis

wapitiCore/net/scope.py
```diff
@@ -46,29 +46,37 @@ def name(self) -> str:
         return self._scope
 
     def check(self, resource: Union[Request, str]) -> bool:
+
+        checked = None
+
         if not resource:
-            return False
+            checked = False
```
devl00p, line 53: I think it should be a `return` here otherwise the scope `punk` may return `True` when a resource is invalid

wapitiCore/net/scope.py
```diff
@@ -46,29 +46,37 @@ def name(self) -> str:
         return self._scope
 
     def check(self, resource: Union[Request, str]) -> bool:
+
+        checked = None
```
devl00p, line 50: I don't see any benefit of having that variable instead of using return statements. If the only purpose is to avoid the linter it seems ok to me to use a `# pylint: disable=R0911` because refactoring doesn't seem coherent here.

If you prefer to keep the variable, protect each branch with an `elif` instead as only one scope can be chosen

tests/web/test_scope.py
```diff
@@ -69,7 +73,19 @@ async def test_scopes():
...
         "http://perdu.com/subdir/page.html?k=v",
-        "http://perdu.com/subdir/page.html"
+        "http://perdu.com/subdir/page.html",
+        "http://subdomain.perdu.com/",
+        "http://subdomain.perdu.com/page.html",
+        "http://subdomain.perdu.com/subdir/subdirpage.html",
+        "https://subdomain.perdu.com/"
+    }
+
+    scope = Scope(Request("http://subdomain.perdu.com/"), "subdomain")
```
devl00p, line 83: Logic seems good to me. Maybe use a full URL to make the test more explicit like what was done for the `domain` scope

`scope = Scope(Request("http://subdomain.perdu.com/subdir/page.html?k=v"), "subdomain")`

wapitiCore/net/scope.py
Darkiros: Yes it is to avoid linter

devl00p (approved):

PR #500, opened 2023-09-22
Add time check for time-consuming modules

bretfourbe:
Add max-attack-time check for time-consuming modules.

Fix https://github.com/wapiti-scanner/wapiti/issues/498

### DIFF
--- a/wapitiCore/attack/mod_buster.py
+++ b/wapitiCore/attack/mod_buster.py
@@ -18,12 +18,13 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 import asyncio
-from typing import Optional
 from os.path import join as path_join
+from time import monotonic
+from typing import Optional
 
 from httpx import RequestError
 
-from wapitiCore.main.log import log_red, log_verbose
+from wapitiCore.main.log import log_red, log_verbose, logging
 from wapitiCore.attack.attack import Attack
 from wapitiCore.net import Request, Response
 from wapitiCore.definitions.buster import NAME, WSTG_CODE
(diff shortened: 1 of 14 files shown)


tarraschk:
Salut @devl00p peux-tu jeter un oeil ?

wapitiCore/attack/mod_buster.py
```diff
@@ -99,6 +100,12 @@ async def test_directory(self, path: str):
 
         with open(path_join(self.DATA_DIR, self.PATHS_FILE), encoding="utf-8", errors="ignore") as wordlist:
             while True:
+                if (datetime.utcnow() - self.start).total_seconds() > self.max_attack_time >= 1:
```
devl00p, line 103: It will definitively work although I think the recommended solution would be to use `time.monotonic` to prevent changes in computer time.

Also it should be possible to leverage the `_stop_event` but it would be more difficult and would need more testing (something like creating a task that will set the event after X seconds, https://stackoverflow.com/questions/45419723/python-timer-with-asyncio-coroutine/45430833#45430833 ).

Maybe let's create another issue for that behavior.

I'm ok with the current MR solution but it requires more code in each module.

devl00p (approved):

wapitiCore/attack/mod_buster.py
tarraschk: @bretfourbe peux-tu faire la modif du `time.monotonic` ? 

Tu peux ignorer la partie `_stop_event` pour le moment, ce serait à faire dans une issue séparée sauf si tu dis que c'est trivial.

bretfourbe: I changed to `time.monotonic` :) 
@devl00p i think only few modules require this change since their execution can take a lot of time for each request

bretfourbe:
I just added tests except for `wp_enum` where i did not succeed in delaying responses to make the module execution slower. @devl00p  if you have any idea ? I tried with `side_effect` to call a function with `await asyncio.sleep(1)` for instance but it raises errors :/

PR #513, opened 2023-10-20
Adding Swagger option

Darkiros:
Adding `--swagger` option to import API routes into Request object with its parameters to perform a scan.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,7 +51,8 @@ dependencies = [
         "sqlalchemy==2.0.28",
         "tld==0.13",
         "typing-extensions==4.10.0",
-        "yaswfp==0.9.3"
+        "yaswfp==0.9.3",
+        "prance[osv]>=23.6.21.0"
 ]
 
 [project.urls]
@@ -69,7 +70,7 @@ ssl = [
 ]
 test = [
         "humanize==4.9.0",
-        "pytest==8.0.2", 
+        "pytest==8.0.2",
         "pytest-cov==4.1.0",
         "pytest-asyncio==0.23.5",
         "respx==0.20.2",
--- a/wapitiCore/parsers/swagger.py
+++ b/wapitiCore/parsers/swagger.py
@@ -1,0 +1,376 @@
+import json
+import sys
+import urllib.parse
+from prance import ResolvingParser, ValidationError
+from prance.util.url import ResolutionError
+
+from wapitiCore.net import Request
+from wapitiCore.main.log import logging
+
+class Swagger():
+    AUTOFILL_VALUES = {
+        "file": ("pix.gif", b"GIF89a", "image/gif"),
+        "integer": "1337",
+        "number": "13.37",
+        "string": "default",
+        "time": "13:37",
+        "url": "https://wapiti-scanner.github.io/",
+        "boolean": "true",
+        "object": {},
+    }
+
+    swagger_dict = None
+
+
+    def __init__(self, swagger_url: str = None, base_url: str = None) -> None:
+        if swagger_url:
+            try:
+                self.swagger_dict = ResolvingParser(swagger_url, backend='openapi-spec-validator',
+                                                strict=False, recursion_limit=5).specification
+            except ValidationError as e:
+                logging.error("[-] Error: Swagger file is not valid : " + str(e.args[0]) +
+                                ". See https://swagger.io/specification/ for more information.")
+                sys.exit(1)
+            except AssertionError:
+                logging.error("[-] Error: File not found")
+                sys.exit(1)
+            except ResolutionError:
+                logging.error("[-] Error: Unable to resolve the swagger file")
+                sys.exit(1)
+        else:
+            logging.error("[-] Error: No URL or file")
+            sys.exit(1)
+
+        self.routes = self._get_routes(self.swagger_dict, swagger_url, base_url)
+
+
+    @staticmethod
+    def _get_base_url(swagger_dict: dict, url: str) -> str:
+        try:
+            parsed_host = urllib.parse.urlparse(url)
+            if 'schemes' not in swagger_dict:
+                # get http or https from url
+                swagger_dict['schemes'] = parsed_host.scheme
+            if 'host' not in swagger_dict:
+                if url:
+                    swagger_dict['host'] = parsed_host.hostname
+                    if parsed_host.port:
+                        swagger_dict['host'] += ":" + str(parsed_host.port)
+                else:
+                    swagger_dict['host'] = ""
+            elif swagger_dict['host'] == "localhost" and url:
+                swagger_dict['host'] = parsed_host.hostname
+            if 'basePath' not in swagger_dict:
+                swagger_dict['basePath'] = ""
+            if 'https' in swagger_dict['schemes']:
+                return 'https://' + swagger_dict['host'] + swagger_dict['basePath']
+            return 'http://' + swagger_dict['host'] + swagger_dict['basePath']
+        except ValueError as e:
+            logging.error("[-] Error: Swagger file is not valid : " + str(e) +
+                          ". See https://swagger.io/specification/ for more information.")
+            sys.exit(1)
+
+
+    @staticmethod
+    def _check_properties(model_name: dict) -> dict:
+        if "properties" in model_name:
+            return model_name['properties']
+        if "additionalProperties" in model_name:
+            return model_name['additionalProperties']
+        return model_name
+
+
+    # Parse object in swagger file.
+    # Replace all object by their type and all array by their type
+    # acording to their properties and definitions.
+    # It will be easier to create request with default value.
+    def _parse_object(self, model_name):
+        try:
+            model = {}
+            for key in model_name:
+                if 'type' in model_name[key]:
+                    if 'object' in model_name[key]['type']:
+                        ref = self._check_properties(model_name[key])
+                        model[key] = self._parse_object(ref)
+                        if 'type' in model[key]:
+                            if model[key]['type'] == "array":
+                                model[key] = {"array": model[key]['items']}
+                            else:
+                                model[key] = model[key]['type']
+                    elif 'array' in model_name[key]['type']:
+                        if 'type' in model_name[key]['items']:
+                            model[key] = {"array": model_name[key]['items']['type']}
+                            if 'object' in model_name[key]['items']['type']:
+                                ref = self._check_properties(model_name[key]['items'])
+                                model[key]["array"] = self._parse_object(ref)
+                    else:
+                        model[key] = model_name[key]['type']
+                else:
+                    model[key] = model_name[key]
+            return model
+        except ValueError as e:
+            logging.error("[-] Error: Swagger file is not valid\n" + str(e) +
+                          "\nSee https://swagger.io/specification/ for more information")
+            sys.exit(1)
+
+
+    def _check_params(self, params: dict) -> list:
+        raws = []
+        for param in params:
+            raw = {}
+            if 'in' in param or '/' in param:
+                if '/' in param:
+                    raw['in'] = "body"
+                    param = params[param]
+                else:
+                    raw['in'] = param['in']
+                if raw['in'] == "body" and 'schema' in param:
+                    if 'type' in param['schema']:
+                        if 'object' in param['schema']['type']:
+                            ref = self._check_properties(param['schema'])
+                            model = self._parse_object(ref)
+                            raw['model'] = model
+                        elif 'array' in param['schema']['type']:
+                            if 'object' in param['schema']['items']['type']:
+                                ref = self._check_properties(param['schema']['items'])
+                                model = self._parse_object(ref)
+                                raw['model'] = model
+                        else:
+                            raw['type'] = param['schema']['type']
+            if 'type' in param:
+                if param['type'] == "array":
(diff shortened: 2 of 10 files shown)


bretfourbe (approved):

pyproject.toml
```diff
@@ -51,7 +51,8 @@ dependencies = [
         "six>=1.15.0",
         "sqlalchemy>=1.4.26",
         "tld>=0.12.5",
-        "yaswfp>=0.9.3" 
+        "yaswfp>=0.9.3",
+        "prance[osv]"
```
Slokilla, line 55: pas de version précisée ?

wapitiCore/parsers/swagger.py
```diff
@@ -0,0 +1,293 @@
...
+                sys.exit(1)
+        else:
+            logging.error("[-] Error: No URL or file")
+            sys.exit(1)
+
+        self.routes = self._get_routes(swaggerFile, uri)
+
+
+    @staticmethod
+    def _get_base_url(swaggerFile: dict, url: str) -> str:
```
devl00p, line 45: maybe rename `swaggerFile` to  `swagger_dict` or `swagger_info` ?

wapitiCore/parsers/swagger.py
```diff
@@ -0,0 +1,293 @@
...
+        try:
+            if 'schemes' not in swaggerFile:
+                # get http or https from url
+                swaggerFile['schemes'] = urllib.parse.urlparse(url).scheme
+            if 'host' not in swaggerFile:
+                if url:
+                    swaggerFile['host'] = urllib.parse.urlparse(url).hostname
```
devl00p, line 52: I don't know how swagger files look but what will happen if the URL is on a non-standard port ?

Maybe we should use `.netloc` instead ?

wapitiCore/parsers/swagger.py
```diff
@@ -0,0 +1,293 @@
...
+                return 'https://' + swaggerFile['host'] + swaggerFile['basePath']
+            return 'http://' + swaggerFile['host'] + swaggerFile['basePath']
+        except ValueError as e:
+            logging.error("[-] Error: Swagger file is not valid : " + str(e) +
+                          ". See https://swagger.io/specification/ for more information.")
+            sys.exit(1)
+
+
+    @staticmethod
+    def _check_properties(model_name):
```
devl00p, line 69: It would great to have annotations for `model_name` here and bellow

devl00p: same goes for the return value even if I can guess it is a dict

wapitiCore/parsers/swagger.py
```diff
@@ -0,0 +1,293 @@
...
+        if "properties" in model_name:
+            return model_name['properties']
+        if "additionalProperties" in model_name:
+            return model_name['additionalProperties']
+        return model_name
+
+
+    def _parse_object(self, model_name):
```
devl00p, line 77: As someone who is missing Swagger knowledge :upside_down_face:  , it would be great to have a description about what that function is doing

wapitiCore/parsers/swagger.py
```diff
@@ -0,0 +1,293 @@
+import json
```
devl00p, line 1: Great job. I think that file needs more explanation but looking at the tests it seems promising :+1: 

I know globally what swagger is but never put my eyes in those json/yml files.

devl00p:
There is a last problem I see there: the `prance` module is using `requests` to fetch an external URL so it doesn't respect the proxy options if set.

Also it uses the `requests` default user-agent meaning Wapiti may be ok to crawl the website but fetching the swagger file may fail if the website is behind cloudflare (which will notice the `requests` UA).

Can you please fix the feature in order to use the Wapiti crawler instance if a swagger URL is given ? We would then feed `prance` which a local copy of the swagger file.

devl00p:
It looks like it doesn't work with parameters having `in` set to `query` and a type defined through the `schema` dictionary.

For example with the following API : https://api.kolibricloud.ch/swagger/index.html?url=/swagger/v1/swagger.json

The swagger.json url is `https://api.kolibricloud.ch/swagger/v1/swagger.json`

It will crash on this entry as it seems to be the first one returned by `prance`:

```json
    "/v1/AdministrationSettings/MailAccount": {                                                                        
      "delete": {                                                                                                      
        "tags": [                                                                                                      
          "AdministrationSettings"                                                                                     
        ],                                                                                                             
        "operationId": "AdministrationSettings_DeleteMailAccount",                                                     
        "parameters": [                                                                                                
          {                                                                                                            
            "name": "id",                                                                                              
            "in": "query",                                                                                             
            "schema": {                                                                                                
              "type": "integer",                                                                                       
              "format": "int32"                                                                                        
            }                                                                                                          
          }                                                                                                            
        ],                                                                                                             
        "responses": {                                                                                                 
          "200": {                                                                                                     
            "description": "Success"                                                                                   
          }                                                                                                            
        }                                                                                                              
      }                                                                                                                
    }
```

It will crash here:

```
  File "wapiti-code/wapitiCore/parsers/swagger.py", line 211, in _transform_query
    if 'enum' in param['type']:
KeyError: 'type'
```

The problem seems to com from https://github.com/wapiti-scanner/wapiti/pull/513/files#diff-7155b6ced8c5a5e4f9c4ad2d302e5b4e904d5f16bfd63f719cf369e86d16c964R120

In `_check_params` it checks for `param['in'] == "body"` so it doesn't enter the `if` block and therefore doesn't extract information from the `schema` key while it is present.

Maybe the `if` statement is not broad enough ? I din't dive into the openapi specs

Darkiros:
> It looks like it doesn't work with parameters having `in` set to `query` and a type defined through the `schema` dictionary.
> 
> For example with the following API : https://api.kolibricloud.ch/swagger/index.html?url=/swagger/v1/swagger.json
> 
> The swagger.json url is `https://api.kolibricloud.ch/swagger/v1/swagger.json`
> 
> It will crash on this entry as it seems to be the first one returned by `prance`:
> 
> ```json
>     "/v1/AdministrationSettings/MailAccount": {                                                                        
>       "delete": {                                                                                                      
>         "tags": [                                                                                                      
>           "AdministrationSettings"                                                                                     
>         ],                                                                                                             
>         "operationId": "AdministrationSettings_DeleteMailAccount",                                                     
>         "parameters": [                                                                                                
>           {                                                                                                            
>             "name": "id",                                                                                              
>             "in": "query",                                                                                             
>             "schema": {                                                                                                
>               "type": "integer",                                                                                       
>               "format": "int32"                                                                                        
>             }                                                                                                          
>           }                                                                                                            
>         ],                                                                                                             
>         "responses": {                                                                                                 
>           "200": {                                                                                                     
>             "description": "Success"                                                                                   
>           }                                                                                                            
>         }                                                                                                              
>       }                                                                                                                
>     }
> ```
> 
> It will crash here:
> 
> ```
>   File "wapiti-code/wapitiCore/parsers/swagger.py", line 211, in _transform_query
>     if 'enum' in param['type']:
> KeyError: 'type'
> ```
> 
> The problem seems to com from https://github.com/wapiti-scanner/wapiti/pull/513/files#diff-7155b6ced8c5a5e4f9c4ad2d302e5b4e904d5f16bfd63f719cf369e86d16c964R120
> 
> In `_check_params` it checks for `param['in'] == "body"` so it doesn't enter the `if` block and therefore doesn't extract information from the `schema` key while it is present.
> 
> Maybe the `if` statement is not broad enough ? I din't dive into the openapi specs

I fixed the problem in the new commit.

devl00p:
This is a lot more stable :)

I tried on several websites I found, unfortunately `prance` is not able to parse everything.

I also noticed a crash with https://prufa.signet.is/SignetService2/v3/swagger/v1/swagger.json

```json
Traceback (most recent call last):
  File "/home/devloop/travaux/wapiti-code/./bin/wapiti", line 34, in <module>
    wapiti_asyncio_wrapper()
  File "/home/devloop/travaux/wapiti-code/wapitiCore/main/wapiti.py", line 432, in wapiti_asyncio_wrapper
    asyncio.run(wapiti_main())
  File "/home/devloop/.pyenv/versions/3.9.15/lib/python3.9/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/home/devloop/.pyenv/versions/3.9.15/lib/python3.9/asyncio/base_events.py", line 647, in run_until_complete
    return future.result()
  File "/home/devloop/travaux/wapiti-code/wapitiCore/main/wapiti.py", line 163, in wapiti_main
    for request in swagger.get_requests():
  File "/home/devloop/travaux/wapiti-code/wapitiCore/parsers/swagger.py", line 359, in get_requests
    return self._create_request(self.routes)
  File "/home/devloop/travaux/wapiti-code/wapitiCore/parsers/swagger.py", line 322, in _create_request
    data = self._transform_body(param)
  File "/home/devloop/travaux/wapiti-code/wapitiCore/parsers/swagger.py", line 285, in _transform_body
    json_dict[key] = self._replace_param(param['model'][key])
  File "/home/devloop/travaux/wapiti-code/wapitiCore/parsers/swagger.py", line 355, in _replace_param
    json_dict[key] = self.AUTOFILL_VALUES[json_dict[key]]
TypeError: unhashable type: 'list'
```

What is hapenning here is that `key` is a `allOf` entry that keeps a list (see [specification](https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/)) of accepted subschemas.

As we can't use a list as a key in a dict, the statement on the right side of the equal sign fails.

If it sounds too complicated to handle such cases I think we should just ignore them. Having a broad exception in `_create_request` to catch complicated routes should be good.

And finally, the error message `Error: Unable to resolve the swagger file` is misleading. Without taking an in-depth look at the code I would never had guess that it is not tied to a network error. I know prance raises a `ResolutionError` but for users we should just tell the swagger file is invalid.

Darkiros:
> This is a lot more stable :)
> 
> I tried on several websites I found, unfortunately `prance` is not able to parse everything.
> 
> I also noticed a crash with https://prufa.signet.is/SignetService2/v3/swagger/v1/swagger.json
> 
> ```json
> Traceback (most recent call last):
>   File "/home/devloop/travaux/wapiti-code/./bin/wapiti", line 34, in <module>
>     wapiti_asyncio_wrapper()
>   File "/home/devloop/travaux/wapiti-code/wapitiCore/main/wapiti.py", line 432, in wapiti_asyncio_wrapper
>     asyncio.run(wapiti_main())
>   File "/home/devloop/.pyenv/versions/3.9.15/lib/python3.9/asyncio/runners.py", line 44, in run
>     return loop.run_until_complete(main)
>   File "/home/devloop/.pyenv/versions/3.9.15/lib/python3.9/asyncio/base_events.py", line 647, in run_until_complete
>     return future.result()
>   File "/home/devloop/travaux/wapiti-code/wapitiCore/main/wapiti.py", line 163, in wapiti_main
>     for request in swagger.get_requests():
>   File "/home/devloop/travaux/wapiti-code/wapitiCore/parsers/swagger.py", line 359, in get_requests
>     return self._create_request(self.routes)
>   File "/home/devloop/travaux/wapiti-code/wapitiCore/parsers/swagger.py", line 322, in _create_request
>     data = self._transform_body(param)
>   File "/home/devloop/travaux/wapiti-code/wapitiCore/parsers/swagger.py", line 285, in _transform_body
>     json_dict[key] = self._replace_param(param['model'][key])
>   File "/home/devloop/travaux/wapiti-code/wapitiCore/parsers/swagger.py", line 355, in _replace_param
>     json_dict[key] = self.AUTOFILL_VALUES[json_dict[key]]
> TypeError: unhashable type: 'list'
> ```
> 
> What is hapenning here is that `key` is a `allOf` entry that keeps a list (see [specification](https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/)) of accepted subschemas.
> 
> As we can't use a list as a key in a dict, the statement on the right side of the equal sign fails.
> 
> If it sounds too complicated to handle such cases I think we should just ignore them. Having a broad exception in `_create_request` to catch complicated routes should be good.
> 
> And finally, the error message `Error: Unable to resolve the swagger file` is misleading. Without taking an in-depth look at the code I would never had guess that it is not tied to a network error. I know prance raises a `ResolutionError` but for users we should just tell the swagger file is invalid.

I created an exception for this type of error.
I also took the opportunity to correct an error on this type of file: http://rackerlabs.github.io/wadl2swagger/openstack.html
The format caused a problem when autocompleting variables in urls

devl00p:
@Darkiros Did you try other python modules than `prance` ?

I had the opportunity to use `SwaggerParser` lately and it seems to handle those recursion issues better (but it is slow as hell to parse files :sob:  )

Darkiros:
@devl00p Currently there aren't many other modules capable of doing this.
Like SwaggerParser and some other parsers, Prance is able to read .yaml which is another format used for Swagger.

I had some troubles with SwaggerParser to do what I do with Prance. I couldn't find another module that could replace Prance.
If you know one wich is able to do the job I'm open 😉

devl00p:
rebased to get up-to-date pyproject.toml file

devl00p:
I merged the MR, using prance should be ok for now but I had the chance to work on swagger files recently and it seems that unfortunately there are no python library out there that are able to parse json / yaml swagger files in a bulletproof way.

As we aren't interested in the format of HTTM responses, just the requests, we may have a benefit to write our own parser for those files. Let's keep that idea for later :)

PR #524, opened 2023-11-15
choices=["url", "page", "folder", "domain", "punk"]

deadjdona:
proper order of options from smallest to largest


Slokilla:
What are your intentions ? There's way too much commits here !

PR #533, opened 2023-11-24
fix max-attack-time (again)

bretfourbe:
Add a wait_for condition and remove all checks inside modules
Default value for max-attack-time is None now


fwininger:
Reste à voir pour les tests unitaires, mais la direction bonne.

bretfourbe:
I removed modules unit tests related to max-attack-time since it is not managed by modules anymore but directly in `attack` method from `Wapiti` class (controller)

PR #544, opened 2023-12-07
Add a WordPress scanner

OussamaBeng:
Add a scan for WordPress to the cms module


bretfourbe:
There is no breaking change here for wp_enum module. The module will also be callable with `-m cms --cms wp`.
This also update common paths to check for spip detection

PR #565, opened 2024-02-20
Add module for network devices

OussamaBeng:
Add a module to detect network devices with version when possible

### DIFF
--- a/wapitiCore/attack/mod_network_device.py
+++ b/wapitiCore/attack/mod_network_device.py
@@ -1,0 +1,97 @@
+import json
+
+from typing import Optional
+from urllib.parse import urljoin
+from bs4 import BeautifulSoup
+from httpx import RequestError
+
+from wapitiCore.main.log import logging
+from wapitiCore.attack.attack import Attack
+from wapitiCore.net import Request
+from wapitiCore.net.response import Response
+from wapitiCore.definitions.fingerprint import NAME as TECHNO_DETECTED, WSTG_CODE
+
+from wapitiCore.main.log import log_blue
+
+MSG_TECHNO_VERSIONED = "{0} {1} detected"
+MSG_NO_UBIKA = "No UBIKA Detected"
+
+
+class ModuleNetworkDevice(Attack):
+    """Base class for detecting version."""
+    name = "network_device"
+    version = ""
+
+    async def check_ubika(self, url):
+        check_list = ['app/monitor/']
+        for item in check_list:
+            full_url = urljoin(url, item)
+            request = Request(full_url, 'GET')
+            try:
+                response: Response = await self.crawler.async_send(request, follow_redirects=True)
+            except RequestError:
+                self.network_errors += 1
+                raise
+            soup = BeautifulSoup(response.content, 'html.parser')
+            title_tag = soup.title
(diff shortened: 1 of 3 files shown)


Slokilla (changes requested):

wapitiCore/attack/mod_network_device.py
```diff
@@ -0,0 +1,83 @@
...
+    name = "network_device"
+    version = ""
+
+    async def check_ubika(self, url):
+        check_list = ['app/monitor/']
+        for item in check_list:
+            request = Request(f'{url}{item}', 'GET')
+            try:
+                response: Response = await self.crawler.async_send(request, follow_redirects=True)
+            except RequestError:
```
Qwarctick, line 30: Your function does not return anything if there is a `RequestError` because catched by the try/except.
Consider returning a False by default or raising the same exception.

wapitiCore/attack/mod_network_device.py
```diff
@@ -0,0 +1,83 @@
...
+                self.network_errors += 1
+            else:
+                soup = BeautifulSoup(response.content, 'html.parser')
+                title_tag = soup.title
+                return response.is_success and title_tag and "UBIKA" in title_tag.text.strip()
+
+    async def get_ubika_version(self, url):
+        version_uri = "app/monitor/api/info/product"
+        request = Request(f'{url}{version_uri}', 'GET')
```
Qwarctick, line 39: You could use urllib.parse here too.

wapitiCore/attack/mod_network_device.py
```diff
@@ -0,0 +1,83 @@
...
+
+class ModuleNetworkDevice(Attack):
+    """Base class for detecting version."""
+    name = "network_device"
+    version = ""
+
+    async def check_ubika(self, url):
+        check_list = ['app/monitor/']
+        for item in check_list:
+            request = Request(f'{url}{item}', 'GET')
```
Qwarctick, line 27: You should use `urllib.parse` (from stdlib) for url joining. It's safer to use than string concatenation (in the case you forget or add a double /)

``` python
>>> import urllib.parse
>>> urllib.parse.urljoin("http://foobar.com", "/api/v2")
'http://foobar.com/api/v2'
>>> urllib.parse.urljoin("http://foobar.com", "api/v2")
'http://foobar.com/api/v2'
```

wapitiCore/attack/mod_network_device.py
```diff
@@ -0,0 +1,83 @@
...
+                soup = BeautifulSoup(response.content, 'html.parser')
+                title_tag = soup.title
+                return response.is_success and title_tag and "UBIKA" in title_tag.text.strip()
+
+    async def get_ubika_version(self, url):
+        version_uri = "app/monitor/api/info/product"
+        request = Request(f'{url}{version_uri}', 'GET')
+        try:
+            response: Response = await self.crawler.async_send(request, follow_redirects=True)
+        except RequestError:
```
Qwarctick, line 42: Same here, your error is catched by the try/except and does not return anything in case of error.

Qwarctick (commented):
You can add docstrings for the module and the methods in addtition.

bretfourbe (approved):

Slokilla (approved):

PR #566, opened 2024-02-25
feat : use --no-cache-dir flag to pip in dockerfiles to save space

Rajpratik71:
using the "--no-cache-dir" flag in pip install, make sure downloaded packages by pip don't cache on the system. This is a best practice that makes sure to fetch from a repo instead of using a local cached one. Further, in the case of Docker Containers, by restricting caching, we can reduce image size. In terms of stats, it depends upon the number of python packages multiplied by their respective size. e.g for heavy packages with a lot of dependencies it reduces a lot by don't cache pip packages.

Further, more detailed information can be found at

https://medium.com/sciforce/strategies-of-docker-images-optimization-2ca9cc5719b6

### DIFF
--- a/Dockerfile.test
+++ b/Dockerfile.test
@@ -11,8 +11,8 @@ RUN apt update \
 
 RUN update-ca-certificates
 RUN python3 -c "import sys; print(sys.version)"
-RUN python3 -m pip install --upgrade pip
-RUN pip3 install -U setuptools --no-cache-dir
+RUN python3 -m pip install --no-cache-dir --upgrade pip
+RUN pip3 install --no-cache-dir -U setuptools
 RUN mkdir /usr/src/app
 
 ENV LANG en_US.UTF-8
@@ -23,5 +23,5 @@ WORKDIR /usr/src/app
 
 COPY . .
 
-RUN pip3 install .[test] --no-cache-dir
+RUN pip3 install --no-cache-dir .[test]
 CMD ["pytest"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@ RUN apt update \
 
 COPY . .
 
-RUN pip3 install . --break-system-packages
+RUN pip3 install --no-cache-dir . --break-system-packages
 
 FROM debian:bookworm-slim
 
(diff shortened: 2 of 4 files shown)


Slokilla:
Hello @Rajpratik71, thanks for contributing. Did you measured the difference of image weight before and after ?

Dockerfile.test
```diff
@@ -11,8 +11,8 @@ RUN apt update \
 
 RUN update-ca-certificates
 RUN python3 -c "import sys; print(sys.version)"
-RUN python3 -m pip install --upgrade pip
-RUN pip3 install -U setuptools --no-cache-dir
+RUN python3 -m pip install --no-cache-dir --upgrade pip
+RUN pip3 install --no-cache-dir -U setuptools --no-cache-dir
```
Slokilla, line 15: You duplicated this

Slokilla (changes requested):

Dockerfile.test
Rajpratik71: oh sorry fixed

Slokilla:
Thanks for the correction. I asked you to measure the delta between with and without your changes because you modified multi-stage dockerfiles. As you modify the `build` stage, I do not think it has any impact on the final image.

Rajpratik71:
> Thanks for the correction. I asked you to measure the delta between with and without your changes because you modified multi-stage dockerfiles. As you modify the `build` stage, I do not think it has any impact on the final image.

In all these "Dockerfile", "--no-cache-dir" will help to reduce the overall size of image as one image is being used for other as base image in some cases. 

I don't explicitly checked the delta in terms of docker image size but once the build workflow is approved and succeeds I can extract the diff

Qwarctick:
Disabling the cache is not useful as images use multi-stage builds.
The pip cache folder is by default in `~/.cache/pip` and is not copied to the final stage.

For the sake of argument, I did the test for the main image (Dockerfile) and there's no difference in size.

The only place it might be useful is for the `Dockerfile.test` image, as there is no multi-stage. But this image is only used locally. So it's not much of a problem.

PR #584, opened 2024-03-31
Fix headless wapp for confidence 0

bretfourbe:
Fix headless wapp for confidence 0 such as js detection for lodash and underscore.js
Add case for regex starting with `\\;version:`

### DIFF
--- a/wapitiCore/attack/mod_wapp.py
+++ b/wapitiCore/attack/mod_wapp.py
@@ -51,12 +51,16 @@ VERSION_REGEX = re.compile(r"\d[\d.]*")
 SCRIPT = (
     "wapiti_results = {};\n"
     "for (var js_tech in wapiti_tests) {\n"
+    "  js_tech_results = [];\n"
     "  for (var i in wapiti_tests[js_tech]) {\n"
     "    try {\n"
-    "      wapiti_results[js_tech] = [String(eval(wapiti_tests[js_tech][i])), wapiti_tests[js_tech][i]]; break;\n"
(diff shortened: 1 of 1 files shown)


wapitiCore/attack/mod_wapp.py
```diff
@@ -440,20 +444,32 @@ async def _detect_applications_headless(self, url: str) -> dict:
...
+                                            final_results[software] = [version]
+                                            break
+                                        if software not in final_results:
+                                            validated = True
+                                            final_results[software] = []
+                                    elif expected_format.startswith(r"\;version:"):
+                                        final_results[software] = [expected_format.split(':')[1]]
+                                    elif isinstance(version, str):
+                                        if r"\;confidence:0" in expected_format:
+                                            waiting_version.append(version)
```
devl00p, line 466: what is the story behind those "waiting_version" ?  values we use as a fallback due to lack of confidence ?

bretfourbe: Sometimes we can detect a version but there is a `confidence:0` in the regex so we are not sure whether the result is relevant or not. Instead of just trash the detected version, we store it in `waiting_version`. If another JS command get a valid result for the current techno, it confirms the presence of the techno. If there is no version already detected we use the one in `waiting_version`.
For instance it is usefull in `underscore.js` and `Lodash` detection where the JS command is the same `_.VERSION`. It allows to detetct only one of them instead of both.

devl00p: Great, thx :)

devl00p (approved):

PR #592, opened 2024-05-06
Updating Cookie Flags Module

OussamaBeng:
Add more information about the cookie


devl00p:
I can see your point with that MR but it seems to me that the results will be referencing requests that are not actually tested and parameter names that aren't even parameters. I think it is more meaningful to keep the current behavior

Issue #553, opened 2024-01-31
Add python 3.12 support


Slokilla:
Actually we cannot support 3.12 because it looks like SSLyze & its dependenciees are not 3.12 compatible.

@devl00p what can we do ?

devl00p:
Indeed it represents strong dependencies for little gain + the ssl module is currently optional.

Using sslscan( https://github.com/rbsec/sslscan ) instead may help as it is a binary that is available for several distributions. We could set the module as mandatory and show a warning if the binary is lacking.

sslscan also allows to export the scan result to XML which can be handy for us.

Right now we can deprecate the ssl module in order to support Python 3.12 then later we can put it back with sslscan support.

devl00p:
Started working on the move to sslscan. It will lost support for several attacks (robot, OpenSSL CCS, 0RTT) and certificate chains but most of the features should be kept

bretfourbe:
Sslyze released a 6.0 version for info, i tried to update it here https://github.com/wapiti-scanner/wapiti/pull/548 but there are other  dependencies that are not python3.12 compatible

devl00p:
fixed

PR #602, opened 2024-06-28
Use sslscan in place of sslyze

devl00p:
Theer is still an issue related to certificate transparency logs.

Can't tell at the moment if crt.sh is broken or if I should do things  differently.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -53,7 +53,7 @@ jobs:
     - name: Install dependencies
       run: |
         sudo apt-get update
-        sudo apt-get install php8.1-cli php8.1-xml -y --no-install-recommends
+        sudo apt-get install php8.1-cli php8.1-xml sslscan -y --no-install-recommends
         python -m pip install --upgrade pip
         pip install -U setuptools
         pip3 install .[test]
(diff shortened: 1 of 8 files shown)


devl00p:
related to https://github.com/wapiti-scanner/wapiti/issues/553

Did a simple check for a pattern in crt.sh responses for CT logs.

bretfourbe:
Hi @devl00p,
I think we should remove `ssl` from `common_modules` here : https://github.com/wapiti-scanner/wapiti/blob/master/wapitiCore/attack/attack.py#L81 since `sslscan` is not installed by default.

devl00p:
> Hi @devl00p, I think we should remove `ssl` from `common_modules` here : https://github.com/wapiti-scanner/wapiti/blob/master/wapitiCore/attack/attack.py#L81 since `sslscan` is not installed by default.

Hello,

No, this is on purpose to encourage users to install the dependency. They wil just get a warning telling them to install sslscan.

Qwarctick:
> No, this is on purpose to encourage users to install the dependency. They wil just get a warning telling them to install sslscan.

I does not work. If sslscan is not installed, subprocess will raise a `FileNotFoundError` and stop the execution. You will not have the correct warning.

Issue #605, opened 2024-07-08
Crash on report generation


devl00p:
Bumped into that one, requires more investigation. Python 3.12.

```
[*] Generating report...
Traceback (most recent call last):
  File "/home/devloop/wapiti/./bin/wapiti", line 34, in <module>
    wapiti_asyncio_wrapper()
  File "/home/devloop/.pyenv/versions/my3124/lib/python3.12/site-packages/wapitiCore/main/wapiti.py", line 496, in wapiti_asyncio_wrapper
    asyncio.run(wapiti_main())
  File "/home/devloop/.pyenv/versions/3.12.4/lib/python3.12/asyncio/runners.py", line 194, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/home/devloop/.pyenv/versions/3.12.4/lib/python3.12/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/devloop/.pyenv/versions/3.12.4/lib/python3.12/asyncio/base_events.py", line 687, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/home/devloop/.pyenv/versions/my3124/lib/python3.12/site-packages/wapitiCore/main/wapiti.py", line 483, in wapiti_main
    await wap.attack(global_stop_event)
  File "/home/devloop/.pyenv/versions/my3124/lib/python3.12/site-packages/wapitiCore/controller/wapiti.py", line 583, in attack
    await self.write_report()
  File "/home/devloop/.pyenv/versions/my3124/lib/python3.12/site-packages/wapitiCore/controller/wapiti.py", line 630, in write_report
    self.report_gen.generate_report(self.output_file)
  File "/home/devloop/.pyenv/versions/my3124/lib/python3.12/site-packages/wapitiCore/report/htmlreportgenerator.py", line 97, in generate_report
    mytemplate.render_unicode(
  File "/home/devloop/.pyenv/versions/my3124/lib/python3.12/site-packages/mako/template.py", line 443, in render_unicode
    return runtime._render(
           ^^^^^^^^^^^^^^^^
  File "/home/devloop/.pyenv/versions/my3124/lib/python3.12/site-packages/mako/runtime.py", line 874, in _render
    _render_context(
[...]

PR #607, opened 2024-07-28
Add specific module for LDAP injection

devl00p:


devl00p:
fix https://github.com/wapiti-scanner/wapiti/issues/605

PR #609, opened 2024-08-09
Add CVE findings

devl00p:
- Download on the fly CVE data from wapiti-scanner/nvd-web-cves/ for a specific software when a version is detected. Data is JSON compressed with lzma (XZ)
- Use python `packaging` library for version comparison
- Adds a new finding called "Vulnerable software" which brings some CVE info (severity and description)
- The most recent CVSS format found is used for each CVE
- List of monitored softwares is hardcoded
- Separate wapp update logic from nvd update logic to ease testing
- Fix some issues present in mod_wapp, notably the categories.json and groups.json files were downloaded 27 times each due to bad indentation
- Fix some tests for mod_wapp, notably some assertions were lacking parenthesis to be effective
- Adds unit and integration tests for the new CVE finding

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,6 +47,7 @@ dependencies = [
         "mako==1.3.2",
         "markupsafe==2.1.5",
         "mitmproxy==10.2.3",
+        "packaging==24.1",
         "pyasn1==0.5.1",
         "six==1.16.0",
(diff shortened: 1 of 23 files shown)


devl00p:
tested with `--update` and on during attack with `config` directory removed

PR #611, opened 2024-08-12
Use HTTPArchive/wappalyzer project for mod_wapp

devl00p:
We are relying on a fork of wappalyzer that was abandonned one year ago.

Some new projects are trying to maintain the detection database.

I suggest we rely on https://github.com/enthec/webappanalyzer

Tested it and changes are as simple as changing the repo url in our code.

### DIFF
--- a/tests/attack/test_mod_wapp.py
+++ b/tests/attack/test_mod_wapp.py
@@ -427,10 +427,10 @@ async def test_multi_detection():
         )
     )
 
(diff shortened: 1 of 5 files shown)


devl00p:
One of our tests fails due to a commit they made: https://github.com/enthec/webappanalyzer/commit/42872180dfdeb638e794b1b2e8ece32bc4fb30d2

We can also consider the project https://github.com/HTTPArchive/wappalyzer

devl00p:
let's use the later

PR #625, opened 2024-10-31
Fix signal handling for Windows compatibility in wapiti.py

Havoc199x:
This pull request addresses a compatibility issue with Windows by modifying the signal handling in wapiti.py. The following changes were made:
- Conditional Signal Handling: Added a platform check to ensure that signal handling is only applied on non-Windows systems, preventing the NotImplementedError that occurs on Windows when trying to use asyncio signal handlers.

Background
- In Wapiti 3.2, attempting to run the program on Windows results in a NotImplementedError due to the loop.add_signal_handler() function, which is unsupported on Windows. This PR ensures compatibility by conditionally adding signal handlers only on systems where it’s supported (e.g., Linux and macOS).

Testing
- The changes were tested on a Windows system to confirm that they prevent the error and allow Wapiti to run as expected.

Impact
- This change improves cross-platform compatibility, enabling Windows users to use Wapiti without encountering runtime errors related to signal handling.

### DIFF
--- a/wapitiCore/main/wapiti.py
+++ b/wapitiCore/main/wapiti.py
@@ -22,6 +22,7 @@ import codecs
 import os
 import signal
 import sys
+import platform
 from importlib import import_module
 from inspect import getdoc
 from sqlite3 import OperationalError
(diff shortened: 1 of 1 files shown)


devl00p:
Hello,

I did some changes since your MR. But I think I found a solution that would satisfy both Linux and Windows platforms, still working on it.

Havoc199x:
> Hello,
> 
> I did some changes since your MR. But I think I found a solution that would satisfy both Linux and Windows platforms, still working on it.

Hi,

Thanks for the update! I’d love to contribute and help out with the solution. Please let me know if there’s anything I can do to assist or if you need help testing.

Looking forward to collaborating on this!

devl00p:
I pushed a new version that will work on windows (installing and running).

I used a basic signal handler too, the behavior is the same than what was already made on Linux.

Going further would be complicated as the prompt to ask the user what to do is a blocking task. While it executes, http requests may timeout. Pushing the prompt within a thread would solve that BUT would lead to output being unreadable as both the lines from the scan and the line from the prompt will be printed as the same time. One solution would be to use some TUI librairies like Textual to solve the issue but it may be hard to implement.

Closing this PR

Issue #619, opened 2024-10-08
Issue with `max-attack-time`: Attack Doesn't Stop After Specified Time


Kundan8000:
I'm running the following command, but it isn't stopping the attack after 1 minute. Instead, it gets stuck, and I think I have to manually move to the next attack by pressing <kbd>CTRL</kbd> + <kbd>C</kbd> and selecting the next attack from the menu.
```shell
wapiti \
  --level 1 \
  -u http://testasp.vulnweb.com \
  -f json \
  -o report.json \
  -m all \
  --scope folder \
  --flush-session \
  -v2 \
  --no-bugreport \
  --max-attack-time 60
```

Kundan8000:
@devl00p, please check. It gets stuck after a certain amount of time and either completes after 24 hours or sometimes takes up to 3 days to finish.

devl00p:
Hello,

Which version are you using ?  Which module was running ?

Kundan8000:
I'm running all modules as you can see in command & I've `Wapiti 3.2.0` installed on my server `Ubuntu 22.04.4 LTS`

Kundan8000:
> Hello,
> 
> Which version are you using ? Which module was running ?

I'm running all modules as you can see in above command & I've `Wapiti 3.2.0` installed on my server `Ubuntu 22.04.4 LTS`

Kundan8000:
Hi @devl00p, Have you found anything why it's getting stuck?

Kundan8000:
Hi @devl00p, Thank you for responding. I’ve found another issue: the `brute_login_form` should automatically detect the login form and attempt brute-force login or SQL injection. Currently, this [site](https://www.demoblaze.com/) is vulnerable to **SQL injection** through the login form, which appears as a modal but wapiti fails to detect that.

Kundan8000:
> The `--max-attack-time` is used to specify a max duration of processing for EACH attach module.
> 
> So if you specify 60 seconds with (for example) all 33 attack modules it will take approximately 30 minutes to execute. This is the expected behavior.
> 
> The CTRL+C behavior is not perfect though as the stop signal is checked after each crawled URL, not its mutations. Depending on the module it can takes quite some time but I'm looking to improve it thanks to asyncio cancellation features

But when I start the scan, all the modules take the specified time except for the `buster` module, which takes too long and gets stuck.

devl00p:
[...]

PR #626, opened 2024-11-10
Remove the need for the asyncio stop Event on attacks

devl00p:
THis MR fixes several issues with the `stop_event` mechanism used in the attack phase of Wapiti.

- delay before stopping: for per-requests attack modules, the `stop_event` is checked after each original request is attacked. It can take a while on requests having a lot of parameters along with modules having a lot of payloads. It can be even worse if there is some latency with the target. The new mechanism instantanously cancel the current task, no matter what is currently executed
- `stop_event` having to be handled differently by some modules: some modules like buster, nikto, etc. aren't using "original" requests therefore they have to deal with the `stop_event` in a particular way. Implementing a new attack module can be challenging for new comers that do not know anything about that `stop_event` mechanism
- attacked_ids not always persisted: while it worked with the `stop_event`, I noticed that the `wait_for` on `load_and_attack` (used for `--max-attack-time`)  doesn't allow to get the `attacked_ids` collected so far if the task reach the max time. Removing the `stop_event` brings a behavior kinda similar to `max-attack-time` so it was necessary to refactor the `load_and_attack` function in order to access the `attacked_ids` even if the attack is interrupted. Modifying a set passed as argument and using a `finally` block to persist them fixed that behavior.

### DIFF
--- a/tests/attack/test_mod_backup.py
+++ b/tests/attack/test_mod_backup.py
@@ -1,4 +1,3 @@
-from asyncio import Event
 from unittest.mock import AsyncMock
 
 import httpx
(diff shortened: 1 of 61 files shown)


devl00p:
should fix some behavior mentionned here: https://github.com/wapiti-scanner/wapiti/issues/619

PR #631, opened 2024-11-30
use an arsenic wheel fixed for python 3.12

devl00p:


devl00p:
Tested install + running with python 3.11 on Linux
Tested install + running with python 3.12 on Windows
Tested with headless browser with Python 1.11 on Linux

PR #638, opened 2025-02-18
Fix false positives on ID command injections attacks

tarraschk:
The rules for the ID command injection attacks were too permissive and would lead to several false positives.

Basically, when trying an injection with the `id` command, Wapiti would check if `id=` was available in the results. However, anything that contained `id=` would lead to a false positive, for instance `uuid=`.

Thus, I changed the rules for the id payload, in order to improve its results.

Also, I changed the comment at line 3, which said that rules were tested with a boolean AND when there were multiple case, but it was a boolean OR since a commit 9 months ago.

I tested the mod on the lab https://portswigger.net/web-security/os-command-injection/lab-simple and it still finds vulnerabilities.

### DIFF
--- a/wapitiCore/attack/mod_exec.py
+++ b/wapitiCore/attack/mod_exec.py
@@ -168,7 +168,7 @@ class ModuleExec(Attack):
                 vuln_info = None
 
                 # No timeout raised, check for patterns in response
-                if any(rule in response.content for rule in payload_info.rules):
+                if any(rule.replace("[SPACE]", " ") in response.content for rule in payload_info.rules):
                     vuln_info = payload_info.description
                     # We reached maximum exploitation for this parameter, don't send more payloads
                     vulnerable_parameter = True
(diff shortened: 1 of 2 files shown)


wapitiCore/attack/mod_exec.py
```diff
@@ -168,7 +168,7 @@ async def attack(self, request: Request, response: Optional[Response] = None):
                 vuln_info = None
 
                 # No timeout raised, check for patterns in response
-                if any(rule in response.content for rule in payload_info.rules):
+                if any(rule.replace("<SPACE>", " ") in response.content for rule in payload_info.rules):
```
devl00p, line 171: You should use `[SPACE]` instead, as other ini files are using square brackets, otherwise it looks good

tarraschk: thanks @devl00p, just updated my code with `[SPACE]` instead of `<SPACE>`

devl00p (approved):

PR #645, opened 2025-04-05
wapiti 3.2.4: support Python 3.13

devl00p:


devl00p:
tested with 3.12 and 3.13

PR #748, opened 2026-02-08
Optimize ssrf

fwininger:
Optimisation du module SSRF par Opus 4.6

### DIFF
--- a/tests/attack/test_mod_ssrf.py
+++ b/tests/attack/test_mod_ssrf.py
@@ -10,6 +10,8 @@ from wapitiCore.net import Request
 from wapitiCore.net.crawler import AsyncCrawler
 from wapitiCore.attack.mod_ssrf import ModuleSsrf, SSRF_INBAND_PAYLOADS
 
+PASSWD_CONTENT = "root:x:0:0:root:/root:/bin/bash\ndaemon:x:1:1:daemon:/usr/sbin:/usr/sbin/nologin"
+
 
 @pytest.mark.asyncio
 @respx.mock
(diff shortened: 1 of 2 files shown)


devl00p:
Thanks for this @fwininger. After reviewing it, I've split off the parts that bring clear value into #792, and I'd suggest closing this one in favor of that.

What #792 keeps:
- the `finish()` hardening (skip a request ID that no longer resolves instead of `raise ValueError` aborting the whole result parsing);
- reporting timeout (`ResourceConsumption`) and HTTP 500 (`InternalError`) anomalies during the attack, matching the convention already used by `mod_file`/`mod_sql`.

What #792 intentionally drops, and why:
- **Local file payloads** (`file:///etc/passwd`, `file://C:\...`) duplicate what `mod_file` already tests — same payloads, same `root:x:0:0` rule.
- **`/etc/networks`** is less reliable than `mod_file`'s `/etc/services` (rule `defined by IANA`, present on both Linux and Windows), and its rules (`loopback`, `link-local`) are prone to false positives.
- **GCP / Azure metadata** endpoints are header-gated (`Metadata-Flavor: Google` / `Metadata: true`). Since we inject a parameter *value*, we can't set those headers on the server's outbound request, so these payloads can't trigger.
- **AWS IMDSv1** is deprecated (IMDSv2 blocks the plain GET without a token header).

So the genuinely SSRF-specific in-band detection would reduce to little that `mod_file` doesn't already cover, while adding request volume and false-positive surface. Happy to revisit in-band SSRF separately if we find an angle that isn't already covered elsewhere.

PR #771, opened 2026-02-13
Improve devcontainer for local development workflow

Qwarctick:
This PR improves Improve the devcontainer for local development workflow

- Match container user UID/GID to host (with the user `wapiti`) to avoid permission issues on mounted files
- Mount host .gitconfig, .ssh, and .gnupg for using git command without extra configuration.
- Use a persistent pip cache volume (wapiti-pip-cache) and a venv in the workspace (in `.venv` dir) so dependencies are not re-downloaded on each start

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,13 +1,20 @@
 FROM python:3.13-slim
 
+# UID/GID de l'utilisateur local (passés au build) pour matcher les permissions du montage
+ARG USER_UID=1000
+ARG USER_GID=1000
+
 ENV DEBIAN_FRONTEND=noninteractive \
     LANG=en_US.UTF-8 \
     PYTHONDONTWRITEBYTECODE=1 \
-    PYTHONUNBUFFERED=1
+    PYTHONUNBUFFERED=1 \
+    PLAYWRIGHT_BROWSERS_PATH=/opt/ms-playwright
 
-# Install required system dependencies
-RUN apt-get update && apt-get install -y \
+# Install required system dependencies (dev tools)
+RUN apt-get update && apt-get install -y --no-install-recommends \
     git \
+    openssh-client \
+    gnupg \
     build-essential \
     ca-certificates \
     curl \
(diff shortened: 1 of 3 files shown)


.devcontainer/Dockerfile
```diff
@@ -17,45 +24,45 @@ RUN apt-get update && apt-get install -y \
     php8.4-cli \
     php8.4-xml \
     sslscan \
+    xz-utils \
+    bzip2 \
```
bretfourbe, line 28: do we still need this since we are using playwright install now ?

devl00p:
Hi @Qwarctick, thanks for the contribution and sorry for the late review! First of all: is this PR still relevant for you / something you'd like to see merged? If so, here are a few issues I'd like to see addressed, as the current state has some pitfalls:

**1. `${localEnv:UID}` / `${localEnv:GID}` will almost always be empty.** `UID` and `GID` are bash shell variables, not exported environment variables, so `localEnv` can't see them. The build args arrive empty and the Dockerfile falls back to 1000 — so the mechanism silently does nothing for users whose UID isn't 1000, which is precisely the case it's meant to solve. Note that VS Code already handles this natively: `updateRemoteUserUID` defaults to `true` on Linux whenever a non-root `remoteUser` is set, and rewrites the container user's UID/GID to match the host. So you can keep the `wapiti` user creation but drop the `ARG`/`args` plumbing entirely.

**2. The `~/.gitconfig` / `~/.ssh` / `~/.gnupg` bind mounts are footguns.** If one of them doesn't exist on the host (a contributor without a GPG setup, for instance), Docker creates the missing source path on the host as a **root-owned directory** — the contributor ends up with a root-owned `~/.gitconfig` *directory* on their machine and git breaks on both sides. Also, mounting `~/.gnupg` read-only defeats its purpose: gpg needs to write lock files and sockets there, so signing won't work anyway. And `${localEnv:HOME}` is unset on Windows hosts. Dev Containers already forwards the host git config and the SSH agent into the container out of the box — I'd rather rely on that than on bind mounts.

**3. The persistent pip cache probably doesn't work.** The volume is mounted at `/home/wapiti/.cache/pip`, but that path doesn't exist in the image, so Docker creates the mount point root-owned. pip, running as `wapiti`, can't write to it and silently disables its cache. A `mkdir -p` + `chown` in the Dockerfile is needed.

**4. Playwright version mismatch risk** (related to the earlier inline comment). The image installs the *system* playwright (latest, as root) and downloads Firefox into `/opt/ms-playwright`, but `postCreateCommand` then installs the playwright pinned by `pyproject.toml` (`>=1.42,<2.0`) into the venv. If the two versions diverge, the venv's playwright will expect a different Firefox revision → "Executable doesn't exist" errors — and since `/opt/ms-playwright` is root-owned, the `wapiti` user can't even re-download it. Running `.venv/bin/playwright install firefox` in `postCreateCommand` (with a user-writable `PLAYWRIGHT_BROWSERS_PATH`) would keep the browser in sync with the installed package.

**Minor points:**
- Comments in the Dockerfile should be in English.
- Hardcoding the full `PATH` in `containerEnv` is brittle (it drops the sbin dirs, among others). Prepending via `remoteEnv` with `${containerEnv:PATH}` is safer.
- `[ -d .venv ] || python3 -m venv .venv` will happily reuse a `.venv` created on the host, which won't work inside the container (different interpreter paths).
- `WORKDIR /workspace` is declared twice in the Dockerfile.

PR #772, opened 2026-02-13
Improve Log4Shell DNS verification performance

hbenazha:
This MR improves the performance of the DNS verification used to confirm a Log4Shell exploitation. The module relies on a DNS lookup to determine whether the payload was actually triggered.

To increase performance, we make these DNS lookups asynchronous. The module switches from dns.resolver to dns.asyncresolver so DNS responses can be awaited and batches the requests to run them asynchronously.

To avoid putting too much pressure on the DNS, a semaphore is introduced to cap the number of concurrent request. In addition, the resolver is instantiated once and reused instead of being recreated for every verification.

Tests were updated to reflect the use of an async DNS resolver.

With these changes, runtime went from 11m 20s down to 2m 25s on https://quotes.toscrape.com/

### DIFF
--- a/tests/attack/test_mod_log4shell.py
+++ b/tests/attack/test_mod_log4shell.py
@@ -7,7 +7,7 @@ from httpx import Response as HttpxResponse
 
 import pytest
 import respx
-from dns.resolver import Resolver
+from dns.asyncresolver import Resolver
 
 from tests import get_mock_open
 from wapitiCore.attack.mod_log4shell import ModuleLog4Shell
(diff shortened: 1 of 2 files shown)


bretfourbe (approved):

devl00p:
Thanks @hbenazha — nice, well-targeted performance fix. 👍

The core issue was real: `_verify_dns` used the **blocking** `dns.resolver` inside `async` code, so every DNS verification stalled the whole event loop — and log4shell fires a lot of them. Switching to `dns.asyncresolver`, reusing a single resolver instance, and gathering the header checks under a `Semaphore(4)` is exactly the right approach. The `parallelize_attacks = True` is consistent with the other active modules.

I reproduced the integration test locally (rebuilt wapiti image + tomcat/unifi/dns-endpoint containers): both assertions pass. The earlier `integrate` failure was CI infra flakiness (7s job, 0 steps) — I re-triggered it and everything is green now.

A couple of non-blocking nits for a future cleanup: the class-level `asyncio.Semaphore` binds to the first event loop it's used on (fine for a single scan, slightly fragile across loops); `DNS_MAX_CONCURRENCY` is hardcoded; and `_verify_header_vulnerability` now delegates to `_verify_headers_vulnerability` (wrapping a single header in throwaway dicts), which is a bit indirect. None of these block the merge.

Merging — thanks again!

PR #803, opened 2026-07-14
fix(cli): don't hang on Ctrl+C during an attack (#802)

### LINKED ISSUE #802: Wapiti hangs on Ctrl+C during an attack (asyncio shutdown deadlock in SQLAlchemy/aiosqlite teardown)
## Summary

Pressing `Ctrl+C` during an **attack** phase and then choosing `q` (quit) or `r` (report) can leave Wapiti **hanging forever** instead of exiting — the process has to be killed manually.

## Root cause

When the user interrupts an attack, `ActiveScanner` cancels the running attack tasks (`_attack_wrapper`). Those tasks still hold aiosqlite connections checked out from the SQLAlchemy pool. During cancellation, SQLAlchemy's aiosqlite dialect tries to tear those connections down through `_terminate_graceful_close()`, which is wrapped in `asyncio.shield()`.

`asyncio.run()`'s shutdown step (`_cancel_all_tasks`) then waits for these shielded close coroutines to finish. With **aiosqlite >= 0.22**, `Connection.close()` awaits an internal worker-thread future that never resolves in this cancelled/shutting-down context, so the gather never completes and the event loop teardown deadlocks.

A pending-task dump at the hang shows the deadlock clearly:

```
--- <Task pending ... _terminate_graceful_close() running at
    sqlalchemy/dialects/sqlite/aiosqlite.py:347>  ->  await self._connection.close()
--- <Task cancelling ... ActiveScanner._attack_wrapper() ...>
... (one pair per attack task that was in flight at Ctrl+C time)
```

## Affected configurations

- Reproduced on Python 3.13 with **aiosqlite 0.22.x** + SQLAlchemy 2.0.5x.
- With aiosqlite 0.21 the same teardown path instead raises `NotImplementedError: terminate_force_close() not implemented by this DBAPI shim` (SQLAlchemy calls `Connection.stop`, which only exists in aiosqlite >= 0.22) but the process still exits. So neither pinned version handles Ctrl+C-during-attack cleanly; 0.22 turns the crash into a hang.

## Reproduction

It is a race: it needs several DB connections checked out at the exact instant of the interrupt, so a manual `Ctrl+C` often misses the window. A concurrency-heavy target with per-request latency plus a precisely-timed `SIGINT` triggers it reliably (a standalone reproducer script is attached to the PR).

```
wapiti -u http://<slow-concurrent-target>/ -m xss --flush-session --scope folder -d 1 -v 2
# Ctrl+C while the attack is in flight -> choose q or r -> hang
```

## Fix

Run `wapiti_main()` through an explicit `asyncio.Runner` and arm a short watchdog around the loop-shutdown step: a clean shutdown disarms it immediately, a deadlocked one lets the watchdog force a graceful process exit. The report and session database are already flushed by the time `wapiti_main()` returns, so nothing is lost.

devl00p:
## Summary

Fixes #802: interrupting an attack with Ctrl+C during an active scan could leave Wapiti hanging forever, forcing the user to kill the process.

## Root cause

On interrupt, attack tasks still hold aiosqlite connections checked out from the SQLAlchemy pool. During cancellation SQLAlchemy's aiosqlite dialect tears them down through `_terminate_graceful_close()`, wrapped in `asyncio.shield()`. With aiosqlite >= 0.22, `Connection.close()` awaits an internal worker-thread future that never resolves in that cancelled context, so the event-loop teardown deadlocks.

## Fix

Three changes:

1. **Monkey-patch SQLAlchemy's aiosqlite termination** — wrap `_terminate_graceful_close()` with a 5 s timeout. If the close hangs, `asyncio.TimeoutError` is raised and `terminate()` falls back to `_terminate_force_close()`.

2. **Replace `asyncio.run()` with manual loop control** — `new_event_loop()` + `run_until_complete()` + a custom `_shutdown()` that cancels pending tasks with `asyncio.wait(tasks, timeout=5.0)`.

3. **Skip bug reports in `_attack_wrapper` after cancellation** — DB errors during cleanup are expected side effects, not real bugs.

## Testing

- `pylint` → 10.00/10.
- New tests cover `_shutdown()`, the timed monkey-patch function, and the early return in `_attack_wrapper`.
- Reproduced the original hang and verified the fix with a standalone reproducer (Python 3.13.13, aiosqlite 0.22.1): before → hang; after → clean exit.

### DIFF
--- a/tests/attack/test_active_scanner.py
+++ b/tests/attack/test_active_scanner.py
@@ -1,3 +1,4 @@
+import asyncio
 import types
 from pathlib import Path
 from typing import Tuple
@@ -447,3 +448,33 @@ async def test_init_attack_modules_sorts_by_priority():
     assert {m.name for m in modules} == {"low", "high"}
     # Ensure they are sorted by PRIORITY (high first, then low)
     assert [m.name for m in modules] == ["high", "low"]
+
+
+@pytest.mark.asyncio
+async def test_attack_wrapper_returns_early_when_task_cancelled():
+    scanner = ActiveScanner(MagicMock(), MagicMock())
+    task = asyncio.create_task(asyncio.sleep(3600))
+    task.cancel()
+    try:
+        await task
+    except asyncio.CancelledError:
+        pass
+    scanner._current_attack_task = task
+    assert task.cancelled()
+
+    scanner._bug_report = True
+    scanner.send_bug_report = AsyncMock()
+
+    attack_module = MagicMock()
+    attack_module.name = "test_module"
+    attack_module.must_attack = AsyncMock(side_effect=Exception("db error during cancellation"))
+    attack_module.attack = AsyncMock()
+
+    request = MagicMock()
+    response = MagicMock()
+    attacked_ids = set()
+    semaphore = asyncio.Semaphore(10)
+
+    await scanner._attack_wrapper(attack_module, request, response, attacked_ids, semaphore)
+
+    scanner.send_bug_report.assert_not_called()
--- a/tests/cli/test_shutdown.py
+++ b/tests/cli/test_shutdown.py
@@ -1,0 +1,75 @@
+import sys
+from unittest import mock
+
+import asyncio
+import pytest
+
+from wapitiCore.main.wapiti import _shutdown, wapiti_asyncio_wrapper
+
+
+class TestShutdown:
+    @pytest.mark.asyncio
+    async def test_shutdown_no_pending_tasks(self):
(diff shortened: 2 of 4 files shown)


devl00p:
Superseded by #804, which drops the SQLAlchemy monkey-patch used here (it targeted the base `AsyncAdapt_terminate` mixin, but the aiosqlite connection subclass overrides `_terminate_graceful_close`, so via MRO the patch never applied to the real code path). #804 uses a library-agnostic threading watchdog instead. Closing in favor of it.

